PR #4, opened 2020-10-08
Add BidirectionalCollection.trimming

karwa:
### Description
Adds a '.trimming' method to BidirectionalCollection, returning a slice with matching elements removed from the start and end.

### Detailed Design

```swift

  /// Returns a `SubSequence` formed by discarding all elements at the start and end of the Collection
  /// which satisfy the given predicate.
  ///
  /// e.g. `[2, 10, 11, 15, 20, 21, 100].trimmimg(where: { $0.isMultiple(of: 2) })` == `[11, 15, 20, 21]`
  ///
  /// - parameters:
  ///    - predicate:  A closure which determines if the element should be omitted from the resulting slice.
  ///
  @inlinable
  public func trimming(where predicate: (Element) throws -> Bool) rethrows -> SubSequence
}
```

### Documentation Plan
No existing guides need updating AFAIK. I haven't written a guide for this feature: is it necessary for every API to include this parallel documentation?

### Test Plan
Tests are included.

### Source Impact
Additive.

### Checklist
- [X] I've added at least one test that validates that my change is working, if appropriate
- [?] I've followed the code style of the rest of the project
- [X] I've read the [Contribution Guidelines](CONTRIBUTING.md)
- [X] I've updated the documentation if necessary

### DIFF
--- a/Guides/Trim.md
+++ b/Guides/Trim.md
@@ -1,0 +1,114 @@
+# Trim
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Trim.swift) |
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/TrimTests.swift)]
+
+Returns a `SubSequence` formed by discarding all elements at the start and end of the collection
+which satisfy the given predicate.
+
+This example uses `trimming(where:)` to get a substring without the white space at the beginning and end of the string.
+
+```swift
+let myString = "   hello, world  "
+print(myString.trimming(where: \.isWhitespace)) // "hello, world"
+
+let results = [2, 10, 11, 15, 20, 21, 100].trimming(where: { $0.isMultiple(of: 2) })
+print(results) // [11, 15, 20, 21]
+```
+
+## Detailed Design
+
+A new method is added to `BidirectionalCollection`:
+
+```swift
+extension BidirectionalCollection {
+
+  public func trimming(where predicate: (Element) throws -> Bool) rethrows -> SubSequence
+}
+```
+
+This method requires `BidirectionalCollection` for an efficient implementation which visits as few elements as possible.
+
+A less-efficient implementation is _possible_ for any `Collection`, which would involve always traversing the
+entire collection. This implementation is not provided, as it would mean developers of generic algorithms who forget
+to add the `BidirectionalCollection` constraint will receive that inefficient implementation:
+
+```swift
+func myAlgorithm<Input>(input: Input) where Input: Collection {
+
+  let trimmedInput = input.trimming(where: { ... }) // Uses least-efficient implementation.
+}
+
+func myAlgorithm2<Input>(input: Input) where Input: BidirectionalCollection {
+
+  let trimmedInput = input.trimming(where: { ... }) // Uses most-efficient implementation.
+}
+```
+
+Swift provides the `BidirectionalCollection` protocol for marking types which support reverse traversal,
+and generic types and algorithms which want to make use of that should add it to their constraints.
+
+### Complexity
+
+Calling this method is O(_n_).
+
+### Naming
+
+The name `trim` has precedent in other programming languages. Another popular alternative might be `strip`.
+
+| Example usage | Languages |
+|-|-|
+| ''String''.Trim([''chars'']) | C#, VB.NET, Windows PowerShell |
+| ''string''.strip(); | D |
+| (.trim ''string'') | Clojure |
+| ''sequence'' [ predicate? ] trim | Factor |
+| (string-trim '(#\Space #\Tab #\Newline) ''string'') | Common Lisp |
+| (string-trim ''string'') | Scheme |
+| ''string''.trim() | Java, JavaScript (1.8.1+), Rust |
+| Trim(''String'') | Pascal, QBasic, Visual Basic, Delphi |
+| ''string''.strip() | Python |
+| strings.Trim(''string'', ''chars'') | Go |
+| LTRIM(RTRIM(''String'')) | Oracle SQL, T-SQL |
+| string:strip(''string'' [,''option'', ''char'']) | Erlang |
+| ''string''.strip or ''string''.lstrip or ''string''.rstrip | Ruby |
+| trim(''string'') | PHP, Raku |
+| [''string'' stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]] | Objective-C/Cocoa |
+| ''string'' withBlanksTrimmed ''string'' withoutSpaces ''string'' withoutSeparators | Smalltalk |
+| string trim ''$string'' | Tcl |
+| TRIM(''string'') or TRIM(ADJUSTL(''string'')) | Fortran |
+| TRIM(''string'') | SQL |
+| String.trim ''string'' | OCaml 4+ |
+
+Note: This is an abbreviated list from Wikipedia. [Full table](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(string_functions)#trim)
+
+The standard library includes a variety of methods which perform similar operations:
+
+- Firstly, there are `dropFirst(Int)` and `dropLast(Int)`. These return slices but do not support user-defined predicates.
+  If the collection's `count` is less than the number of elements to drop, they return an empty slice.
+- Secondly, there is `drop(while:)`, which also returns a slice and is equivalent to a 'left-trim' (trimming from the head but not the tail).
+  If the entire collection is dropped, this method returns an empty slice.
+- Thirdly, there are `removeFirst(Int)` and `removeLast(Int)` which do not return slices and actually mutate the collection.
+  If the collection's `count` is less than the number of elements to remove, this method triggers a runtime error.
+- Lastly, there are the `popFirst()` and `popLast()` methods, which work like `removeFirst()` and `removeLast()`,
+  except they do not trigger a runtime error for empty collections.
+
+The closest neighbours to this function would be the `drop` family of methods. Unfortunately, unlike `dropFirst(Int)`,
+the name `drop(while:)` does not specify which end(s) of the collection it operates on. Moreover, one could easily
+mistake code such as:
+
+```swift
+let result = myString.drop(while: \.isWhitespace)
+```
+
+With a lazy filter that drops _all_ whitespace characters regardless of where they are in the string.
+Besides that, the root `trim` leads to clearer, more conscise code, which is more aligned with other programming
+languages:
+
+```swift
+// Does `result` contain the input, trimmed of certain elements?
+// Or does this code mutate `input` in-place and return the elements which were dropped?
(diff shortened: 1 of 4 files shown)


karwa:
In terms of code style, I think my defaults match this project (except that I use a 120 line length).

Does this project just use the default swift-format settings?

Sources/Algorithms/Trim.swift
```diff
@@ -0,0 +1,40 @@
...
+  /// - parameters:
+  ///    - predicate:  A closure which determines if the element should be omitted from the resulting slice.
+  ///
+  @inlinable
+  public func trimmed(where predicate: (Element) throws -> Bool) rethrows -> SubSequence {
+    var sliceStart = startIndex
+    var sliceEnd = endIndex
+    // Consume elements from the front.
+    while sliceStart != sliceEnd, try predicate(self[sliceStart]) {
+      sliceStart = index(after: sliceStart)
```
timvermeulen, line 28:
```suggestion
      formIndex(after: &sliceStart)
```

Sources/Algorithms/Trim.swift
```diff
@@ -0,0 +1,40 @@
...
+  ///
+  /// - parameters:
+  ///    - predicate:  A closure which determines if the element should be omitted from the resulting slice.
+  ///
+  @inlinable
+  public func trimmed(where predicate: (Element) throws -> Bool) rethrows -> SubSequence {
+    var sliceStart = startIndex
+    var sliceEnd = endIndex
+    // Consume elements from the front.
+    while sliceStart != sliceEnd, try predicate(self[sliceStart]) {
```
natecook1000, line 27: Can you use existing stdlib methods here, instead of searching for the start and end manually?

karwa: I can use `firstIndex(where:)` to find the start, but not `lastIndex(where:)` to find the end (since we actually want the index _after_ the last index to match the predicate; this would mean an extra call to `index(after:)`).

I've tried to improve the documentation, too. It might be worth adding a `String.trimmed()` function which defaults to trimming whitespace.

karwa:
@natecook1000 I've added a guide document. It may be a little verbose, I don't know 😅. That's just my style - it's how I write, but I did my best.

I agree with `trim` being a better root verb -- or `trimmed` as an adjective, which reads like "this is the collection, trimmed of these elements", with the verb form perhaps replacing/expanding on the mutating `removeFirst/Last(Int)` methods. I've mentioned the various methods in the standard library in the guide document. I tried not to get too bogged-down in that discussion, otherwise it'd turn more in to a swift-evolution pitch than a useful document for people using this library.

karwa:
@natecook1000 is there anything remaining for me to do here?

Guides/Trim.md
```diff
@@ -0,0 +1,116 @@
...
+
+  public func trimmed(where predicate: (Element) throws -> Bool) rethrows -> SubSequence
+}
+```
+
+This method requires `BidirectionalCollection` for an efficient implementation which visits as few elements as possible.
+
+A less-efficient implementation is _possible_ for any `Collection`, which would involve always traversing the
+entire collection. This implementation is not provided, as it would mean developers of generic algorithms who forget
+to add the `BidirectionalCollection` constraint will recieve that inefficient implementation:
```
natecook1000, line 36: Agreed, `BidirectionalCollection` is the right place for this.

Sources/Algorithms/Trim.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+extension BidirectionalCollection {
+
+  /// Returns a `SubSequence` formed by discarding all elements at the start and end of the Collection
+  /// which satisfy the given predicate.
+  ///
+  /// Trimming is a common operation in computing, often referring to removing leading and trailing whitespace
+  /// from a string[1].
```
natecook1000, line 18: I don't think we need this explanation. Let's introduce the example, instead — something like:

> This example uses `trimmed(where:)` to get a substring without the white space at the beginning and end of the string.

Sources/Algorithms/Trim.swift
```diff
@@ -0,0 +1,48 @@
...
+  ///
+  ///  ```
+  ///  let myString = "  hello, world  "
+  ///  print(myString.trimmed(where: \.isWhitespace)) // "hello, world"
+  ///  ```
+  ///
+  /// [1]: https://en.wikipedia.org/wiki/Trimming_(computer_programming)
```
natecook1000, line 25:
```suggestion
```

We should mention that if the entire collection is trimmed, the  result is an empty subsequence at the end of the collection.

natecook1000 (commented):
Looking good! Just a couple notes about the documentation.

Guides/Trim.md
xwu, line 36:
```suggestion
to add the `BidirectionalCollection` constraint will receive that inefficient implementation:
```

Guides/Trim.md
```diff
@@ -0,0 +1,116 @@
...
+  let trimmedInput = input.trimmed(where: { ... }) // Uses most-efficient implementation.
+}
+```
+
+Swift provides the `BidirectionalCollection` protocol for marking types which support reverse traversal,
+and generic types and algorithms which want to make use of that should add it to their constraints.
+
+### Complexity
+
+Calling this methods is O(_n_).
```
xwu, line 55:
```suggestion
Calling this method is O(_n_).
```

Sources/Algorithms/Trim.swift
```diff
@@ -0,0 +1,48 @@
...
+  public func trimmed(where predicate: (Element) throws -> Bool) rethrows -> SubSequence {
+
+    // Consume elements from the front.
+    let sliceStart = try firstIndex(where: { try predicate($0) == false }) ?? endIndex
+    // sliceEnd is the index _after_ the last index to match the predicate.
+    var sliceEnd = endIndex
+    while sliceStart != sliceEnd {
+      let idxBeforeSliceEnd = index(before: sliceEnd)
+      guard try predicate(self[idxBeforeSliceEnd]) else {
+        return self[sliceStart..<sliceEnd]
```
xwu, line 42: Q: can these bounds be safely unchecked too?

Sources/Algorithms/Trim.swift
```diff
@@ -0,0 +1,48 @@
...
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension BidirectionalCollection {
+
+  /// Returns a `SubSequence` formed by discarding all elements at the start and end of the Collection
```
xwu, line 14: Nit: I _think_ this project's style is to wrap to 80 columns.

Guides/Trim.md
```diff
@@ -0,0 +1,116 @@
+# Trim
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Trim.swift) |
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/TrimTests.swift)]
+
+Returns a `SubSequence` formed by discarding all elements at the start and end of the Collection
+which satisfy the given predicate.
+
+```swift
+let results = [2, 10, 11, 15, 20, 21, 100].trimmed(where: { $0.isMultiple(of: 2) })
```
xwu, line 10: Just a teeny bit of bikeshedding. Because trailing closure syntax causes `where:` to be omitted, there can be a little bit of ambiguity as to whether the closure determine's what's included or what's excluded. It so happens that, in English, "trimming" can take a direct object, while "trimmed" cannot, such that `[1, 2, 3].trimming { $0 == 1 }` suggests that every element equal to one is being dropped a little more strongly than "trimmed" would.

Guides/Trim.md
```diff
@@ -0,0 +1,116 @@
+# Trim
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Trim.swift) |
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/TrimTests.swift)]
+
+Returns a `SubSequence` formed by discarding all elements at the start and end of the Collection
```
xwu, line 6: I think the prevailing style is to use code voice to refer to a `Thing` or otherwise to refer to the thing being modeled in lowercase. A `Collection` is a protocol for collections; a `String` models a string.

```suggestion
Returns a `SubSequence` formed by discarding all elements at the start and end of the collection
```

Guides/Trim.md
```diff
@@ -0,0 +1,116 @@
...
+
+```swift
+let results = [2, 10, 11, 15, 20, 21, 100].trimmed(where: { $0.isMultiple(of: 2) })
+print(results) // [11, 15, 20, 21]
+
+let myString = "   hello, world  "
+print(myString.trimmed(where: \.isWhitespace)) // "hello, world"
+```
+
+Trimming is a common operation in computing. It is particularly for String processing, but also useful in
```
xwu, line 17:
```suggestion
Trimming is a common operation in computing. It is particularly for string processing, but also useful in
```

Sources/Algorithms/Trim.swift
xwu, line 14:
```suggestion
  /// Returns a `SubSequence` formed by discarding all elements at the start and end of the collection
```

xwu (commented):
A remark on documentation.

Sources/Algorithms/Trim.swift
karwa: Personally I'd prefer to leave the position of the empty slice unspecified. If we decided we wanted to change this to first trim elements from the tail and then the head, the result when trimming the entire collection would be an empty subsequence at `startIndex`.

Sources/Algorithms/Trim.swift
karwa: I'm not 100% sure. Since `sliceEnd` is formed by walking backwards from `endIndex` and has not yet tested equal to `sliceStart`, the only way this bounds check will fail is if either your collection or the index's `Comparable` conformance is broken.

For the empty result, we already checked that `sliceStart == sliceEnd` as part of the `while` loop condition, so it is safe to omit the bounds check. A broken conformance would never get to that part.

Sources/Algorithms/Trim.swift
karwa: I ran this through swift-format (the project doesn't have a format file, so I assume it's the default settings?), but it doesn't seem to wrap comments. Will fix.

Guides/Trim.md
karwa: Yeah, that sounds good. I like `trimmed`.

karwa:
Apologies for the delay. I've addressed your comments @xwu and @natecook1000 - documentation has been improved, and I think Xiaodi makes a good case for the name `trimming`. Code in the tests reads quite nicely after this, e.g.:

```
let results = [2, 10, 11, 15, 20, 21, 100].trimming { $0.isMultiple(of: 2) }
```

PR #10, opened 2020-10-09
Fix mistake where `perm` was used in example instead of `combo`

AustinConlon:


natecook1000:
Thanks, @AustinConlon!

PR #11, opened 2020-10-09
Clean up early exit checks

Saklad5:
Early exits using if statements have been changed to use guard where appropriate.

Redundant early exits, where the statement is immediately followed by an equivalent statement with the same behavior, have been removed.

MutableCollection.stablePartition(count:, subrange:, by:) has been restructured to use a switch statement, since a substantial portion of the function consisted of consecutive if statements comparing the same parameter.

This pull request cleans up some of the early exits in the algorithms, either by changing them to use guard or by removing them entirely if they seem clearly redundant.

I’m new to this project, so I cannot rule out the possibility of errors. I ran the unit tests and checked everything manually, but a regression may have been introduced even by changes as simple as this.

I have also not run performance profiling to confirm my changes have no negative impact.

Finally, as there is no official code style for this project, I cannot be sure that I have followed it. The existing code seems to follow normal Swift library conventions, but I may have overlooked something.

### Checklist

### DIFF
--- a/Sources/Algorithms/Partition.swift
+++ b/Sources/Algorithms/Partition.swift
@@ -26,23 +26,23 @@ extension MutableCollection {
     subrange: Range<Index>,
     by belongsInSecondPartition: (Element) throws-> Bool
   ) rethrows -> Index {
-    if n == 0 { return subrange.lowerBound }
-    if n == 1 {
-      return try belongsInSecondPartition(self[subrange.lowerBound])
-        ? subrange.lowerBound
-        : subrange.upperBound
+    switch n {
+    case 0,
+         1 where try belongsInSecondPartition(self[subrange.lowerBound]):
+      return subrange.lowerBound
+    case 1: return subrange.upperBound
+    default:
+      let h = n / 2, i = index(subrange.lowerBound, offsetBy: h)
+      let j = try stablePartition(
+        count: h,
+        subrange: subrange.lowerBound..<i,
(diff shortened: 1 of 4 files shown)


Sources/Algorithms/Partition.swift
```diff
@@ -26,23 +26,23 @@ extension MutableCollection {
...
     by belongsInSecondPartition: (Element) throws-> Bool
   ) rethrows -> Index {
-    if n == 0 { return subrange.lowerBound }
-    if n == 1 {
-      return try belongsInSecondPartition(self[subrange.lowerBound])
-        ? subrange.lowerBound
-        : subrange.upperBound
+    switch n {
+    case 0,
+         1 where try belongsInSecondPartition(self[subrange.lowerBound]):
```
rockbruno, line 31: I personally think this switch is really hard to read. If we don't want the dangling ifs then I think guards could be a better option.

natecook1000: I think this combination of the `0` and `1` cases is the issue for readability. The switch can stay, but let's move the `belongsInSecondPartition` test into the `1` branch.

Saklad5: Fair enough.

natecook1000:
Closing, as we've decided not to take PRs for this kind of refactoring. Feel free to check the issues for other ideas of how to contribute. Thanks!

PR #14, opened 2020-10-10
Added a heap view to Collection that manages a simple Heap structure

dfsweeney:
Added `Heap` to provide a comparator-based in-order view of a `Collection`. 

I know @rakaramos has been working on this with `PartialSort` and some other things that are more in line with the
evolution process. So this is kind of a foul but I put this together to scratch an itch and wanted to sent the PR for people to look at. 

I used the `Permutations` code as a base. There's been some discussion of algorithms vs data structures in the forums. This is a weird one because it includes a data structure to support a partial sort algorithm. 

I'm also not sure if there is not a "currency type" hiding in there around the array of indexes that is also treated as a binary tree. C++ has some methods in `algorithm` that let you treat an array as a binary tree, and the heap algorithms do something like that. 

I put some other notes in `Guides/Heap.md` as well. There are still some to-dos at the top of `Heap.swift`. 

Thanks for looking at this! 

### Checklist

### DIFF
--- a/Guides/Heap.md
+++ b/Guides/Heap.md
@@ -1,0 +1,96 @@
+# Heap
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Heap.swift) | 
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/AlgorithmsTests/HeapTests.swift)]
+
+A type that provides a heap view of a collection’s elements, or of a subset of those 
+elements. 
+
+The heap view puts the maximum element (defined by the provided comparator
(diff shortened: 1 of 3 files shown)


natecook1000:
Thanks for this contribution, @dfsweeney. You're right that the `Heap` data structure belongs in a different package than `Algorithms`. I'm going to close this PR — please feel free to start a thread in the forums if you'd like to continue the discussion. 

My hunch is that you'll ultimately end up with a better data structure actually storing the elements in the internal array, rather than the indices. This kind of wrapper works best when it isn't making any extra allocations, but since you have to copy the indices, you might as well copy the elements instead. Storing the elements would let you implement the other half of the heap operations, as well.

PR #20, opened 2020-10-12
Add sliding windows algorithm

ollieatkinson:
Many thanks to @timvermeulen for helping me refine this collection algorithm.

## Description

Break a collection into overlapping contiguous window subsequences where
elements are slices from the original collection.

The `slidingWindows(ofCount:)` method takes in a integer size and returns a collection of subsequences.

```swift
let swift = "swift"

let windowed = swift.slidingWindows(ofCount: 2) 
// windowed == [ "sw", "wi", "if", "ft" ]
```

## Detailed Design

The `slidingWindows(ofCount:)` is added as a method on an extension of  `Collection`

```swift
extension Collection {
  public func slidingWindows(ofCount count: Int) -> SlidingWindows<Self> {
    SlidingWindows(base: self, size: count)
  }
}
```

If a size larger than the collection length is specified, an empty collection is returned. Due to this 
behaviour the indexes must be calculated on initialisation as we have to be able to compare the 
`upperBound` and allow `Collection` correctly calculate `isEmpty`  .

```swift
[1, 2, 3].slidingWindows(ofCount: 5).isEmpty // true
```

The resulting `SlidingWindows` type is a collection, with conditional conformance to the 
`BidirectionalCollection`, and `RandomAccessCollection`  when the base collection
conforms.

### Complexity

The call to `slidingWindows(ofCount: k)` is O(_1_) if the collection conforms to 
`RandomAccessCollection`, otherwise O(_k_). Access to the next window 
is O(_1_).

### Naming

The type `SlidingWindows` takes its name from the algorithm, similarly the method takes
it's name from it too  `slidingWindows(ofCount: k)`. 

The label on the method `ofCount` was chosen to create a consistent feel to the API 
available in swift-algorithms repository. Inspiration was taken from 
`combinations(ofCount:)` and  `permutations(ofCount:)`.

Previously the name `windows` was considered but was deemed to potentially create 
ambiguity with the Windows operating system.

[rust](https://doc.rust-lang.org/std/slice/struct.Windows.html) has `std::slice::Windows`  which is a method available on slices. It has the same semantics as 
described here.

### Documentation

This documentation exists in `Guides/SlidingWindows.md`

### Impact

This is an additive change, no impact.

---

### Checklist

### DIFF
--- a/Guides/Windows.md
+++ b/Guides/Windows.md
@@ -42,7 +42,8 @@ conforms.
 
 ### Complexity
 
-The call to `windows(size: k)` is O(_k_) and access to the next window is O(_1_).
+The call to `windows(size: k)` is O(_k_) and access to the next window is O(_1_). If the base 
+collection conforms to `RandomAccessCollection` then both are O(_1_).
 
 ### Naming
 
--- a/Guides/Windows.md
+++ b/Guides/Windows.md
@@ -29,15 +29,20 @@ extension Collection {
 
 If a size larger than the collection length is specified, an empty collection is returned. Due to this 
 behaviour the indexes must be calculated on initialisation as we have to be able to compare the 
-`upperBound` in each window slice.
+`upperBound` and allow `Collection` correctly calculate `isEmpty`  .
 
 ```swift
 [1, 2, 3].windows(size: 5).isEmpty // true
 ```
 
+The resulting `Windows` type is a collection, with conditional conformance to the 
+`BidirectionalCollection`, and `RandomAccessCollection`  when the base collection
+conforms.
+
 ### Complexity
 
-O(_N_) time and O(_1_) space complexity.
+The algorithm is O(_N_) time and O(_1_) space complexity.
+Calling `[1, 2, 3].windows(size: k)` would result in O(_k_). 
 
 ### Naming
 
--- a/Guides/Windows.md
+++ b/Guides/Windows.md
@@ -6,7 +6,8 @@
 Break a collection into overlapping contiguous window subsequences where
 elements are slices from the original collection.
 
-The `windows(size:)` method takes in a integer size and returns a collection of subsequences.
+The `windows(size:)` method takes in a integer size and returns a collection of 
+subsequences.
 
 ```swift
 let swift = "swift"
@@ -27,9 +28,9 @@ extension Collection {
 }
 ```
 
-If a size larger than the collection length is specified, an empty collection is returned. Due to this 
-behaviour the indexes must be calculated on initialisation as we have to be able to compare the 
-`upperBound` and allow `Collection` correctly calculate `isEmpty`  .
+If a size larger than the collection length is specified, an empty collection is returned. 
+The first upper bound is computed eagerly because it determines if the collection 
+`startIndex` returns `endIndex`. 
 
 ```swift
 [1, 2, 3].windows(size: 5).isEmpty // true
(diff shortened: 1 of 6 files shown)


Guides/Windows.md
```diff
@@ -0,0 +1,57 @@
...
+[1, 2, 3].windows(size: 5).isEmpty // true
+```
+
+The resulting `Windows` type is a collection, with conditional conformance to the 
+`BidirectionalCollection`, and `RandomAccessCollection`  when the base collection
+conforms.
+
+### Complexity
+
+The call to `windows(size: k)` is O(_k_) and access to the next window is O(_1_).
```
timvermeulen, line 45: We should also mention here that the `windows(size:)` call is O(1) for `RandomAccessCollection`s (see [`partitioningIndex`](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Partition.swift#L161-L199) for something similar).

ollieatkinson: Yes, it's worth mentioning. I like the wording you linked I've taken that and phrased that for this algorithm in https://github.com/apple/swift-algorithms/pull/20/commits/47214b6ce96e20ec6696825557dfa307ef57e15d

kylemacomber:
What do you think about `windows(ofCount:)` rather than `windows(size:)`?

When we were naming `combinations(ofCount:)`, `permutations(ofCount:)`, and `randomSample(count:)`, we tried to consistently use "count" when it referred to the number of elements in a sequence or collection—avoiding synonyms `size` and `length`.

ollieatkinson:
> What do you think about `windows(ofCount:)` rather than `windows(size:)`?
> 
> When we were naming `combinations(ofCount:)`, `permutations(ofCount:)`, and `randomSample(count:)`, we tried to consistently use "count" when it referred to the number of elements in a sequence or collection—avoiding synonyms `size` and `length`.

I would not be against the name change, it was not immediately obvious to me that `ofCount` was an option. I prefer `size` but I also prefer consistent API naming. 

The arguments you make are valid, so I would be happy to change - I would like to see what other folks thoughts are before changing the label.

natecook1000 (commented):
Thanks for this contribution, @ollieatkinson! 👏

ollieatkinson:
@kylemacomber I've given this some thought and renamed the label to `ofCount`

Tests/SwiftAlgorithmsTests/WindowsTests.swift
```diff
@@ -0,0 +1,88 @@
...
+    
+    XCTAssertEqualSequences(w[i], "sw")
+    w.formIndex(after: &i)
+    XCTAssertEqualSequences(w[i], "wi")
+    w.formIndex(after: &i)
+    XCTAssertEqualSequences(w[i], "if")
+    w.formIndex(after: &i)
+    XCTAssertEqualSequences(w[i], "ft")
+    w.formIndex(after: &i)
+    XCTAssertTrue(w[i].isEmpty)
```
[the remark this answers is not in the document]
natecook1000: `Array` is kind of singular, in that it's built so that those index offsets optimize down to simple additions and subtractions. Most of the rest of our standard library collections (like `String`, `Dictionary`, and `Set`) will trap if you try to move an index out of bounds, to enforce the preconditions on the `Collection` protocol requirements.

xwu:
In a general-purpose language that now has Windows support, I would be wary of a general-purpose API vending a type named `Windows`. I do think `SlidingWindows` would be more apt here.

ollieatkinson:
> In a general-purpose language that now has Windows support, I would be wary of a general-purpose API vending a type named `Windows`. I do think `SlidingWindows` would be more apt here.

That's a good thought, I'm liking the suggestion you've made! Thank you @xwu

pyrtsa:
Any reason not to rename the function accordingly now, `windows(ofCount:)` → `slidingWindows(ofCount:)`?

ollieatkinson:
> Any reason not to rename the function accordingly now, `windows(ofCount:)` → `slidingWindows(ofCount:)`?

I'm in the process of renaming the type, function and updating the documentation now!

natecook1000 (approved):
Looks great, @ollieatkinson!

Danappelxx:
Should probably add a link to `Guides/SlidingWindows.md` to the README :)

PR #22, opened 2020-10-13
Fix and improve `Chain`'s index offsetting logic

timvermeulen:
`Chain`'s index offsetting logic is untested and contains a few bugs, all these assertions fail:

```swift
let chain = "abc".chained(with: "XYZ")

do {
  let i = chain.index(chain.startIndex, offsetBy: 3, limitedBy: chain.startIndex)
  XCTAssertNil(i)
}

do {
  let i = chain.index(chain.startIndex, offsetBy: 4)
  let j = chain.index(i, offsetBy: -2)
  XCTAssertEqual(chain[j], "c")
}

do {
  let i = chain.index(chain.startIndex, offsetBy: 3)
  let j = chain.index(i, offsetBy: -1, limitedBy: i)
  XCTAssertNil(j)
}
```

I've put the fixes for these in a separate commit for reference.

Another issue with the current implementation is that `chain.index(i, offsetBy: n)` (with `i` pointing into `base1`) always computes the distance from `i` to `base1.endIndex`, regardless of `n`. As a result, this code traverses the entire string:
```swift
let chain = someVeryLongString().chained(with: [])
let n = someVerySmallNumber()
let _ = chain.index(chain.startIndex, offsetBy: n)
```

This PR fixes this by first calling `base1.index(i, offsetBy: n, limitedBy: base1.endIndex)` before potentially computing the distance to `base1.endIndex`.

This does still mean that when crossing the boundary between `base1` and `base2`, the suffix of `base1` is traversed twice. If we could determine whether or not `Base1: RandomAccessCollection` then we could conditionally replace this logic by repeated `base1.index(after:)` calls, but I'm not sure if that's possible...

### Checklist

### DIFF
--- a/Sources/Algorithms/Chain.swift
+++ b/Sources/Algorithms/Chain.swift
@@ -149,7 +149,7 @@ extension Chain: Collection where Base1: Collection, Base2: Collection {
       } else {
         // The limit only has an effect here if it's "above" `i`
         if limit >= i {
-          return Index(first: limit)
+          return nil
         } else {
           return Index(
             second: base2.index(base2.startIndex, offsetBy: n - d))
@@ -193,7 +193,7 @@ extension Chain: Collection where Base1: Collection, Base2: Collection {
       if n <= d {
         return Index(second: base2.index(i, offsetBy: -n))
       } else {
-        return base1.index(base1.endIndex, offsetBy: -n - d, limitedBy: limit)
+        return base1.index(base1.endIndex, offsetBy: -(n - d), limitedBy: limit)
           .map(Index.init(first:))
       }
 
@@ -203,8 +203,12 @@ extension Chain: Collection where Base1: Collection, Base2: Collection {
         return base2.index(i, offsetBy: -n, limitedBy: limit)
           .map(Index.init(second:))
       } else {
-        return Index(
-          first: base1.index(base1.endIndex, offsetBy: -n - d))
+        // The limit only has an effect here if it's "below" `i`
+        if limit <= i {
+          return nil
+        } else {
+          return Index(first: base1.index(base1.endIndex, offsetBy: -(n - d)))
+        }
       }
     }
   }
--- a/Sources/Algorithms/Chain.swift
+++ b/Sources/Algorithms/Chain.swift
@@ -84,13 +84,17 @@ extension Chain: Collection where Base1: Collection, Base2: Collection {
       }
     }
   }
+  
+  /// Converts an index of `Base1` to the corresponding `Index` by mapping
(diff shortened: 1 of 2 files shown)


Sources/Algorithms/Chain.swift
```diff
@@ -142,27 +143,27 @@ extension Chain: Collection where Base1: Collection, Base2: Collection {
   ) -> Index? {
     switch (i.position, limit.position) {
     case let (.first(i), .first(limit)):
-      let d = base1.distance(from: i, to: base1.endIndex)
-      if n < d {
+      if limit >= i {
+        // `limit` is relevant, so `base2` cannot be reached
```
natecook1000, line 147: Good observation!

Sources/Algorithms/Chain.swift
```diff
@@ -142,27 +143,27 @@ extension Chain: Collection where Base1: Collection, Base2: Collection {
...
         return base1.index(i, offsetBy: n, limitedBy: limit)
           .map(Index.init(first:))
+      } else if let j = base1.index(i, offsetBy: n, limitedBy: base1.endIndex) {
```
natecook1000, line 150: I wish there was an operation that combined this calculation and finding `d` in the next branch, but no luck, right?

natecook1000 (approved):
Looks great, @timvermeulen — thank you! 👏

Sources/Algorithms/Chain.swift
timvermeulen: I agree 100% to the extent that I added [the equivalent](https://github.com/rust-lang/rust/pull/76909) for Rust's `Iterator` recently, specifically to speed up `Chain` 🙂 This would make for a good optimization point if we ever get the chance.

The next best thing would be to write a regular function for this which somehow detects whether the collection supports random-access, is there any hope of that being possible to implement? I assume that info does exist at runtime but I haven't found a way to extract it.

natecook1000: > I agree 100% to the extent that I added [the equivalent](https://github.com/rust-lang/rust/pull/76909) for Rust's `Iterator` recently, specifically to speed up `Chain` 🙂 This would make for a good optimization point if we ever get the chance.

Exactly. I think the only place we have anything like that is in the innards of capturing a sequence in an array, where we return the buffer along with in-progress iterator.
 
> The next best thing would be to write a regular function for this which somehow detects whether the collection supports random-access, is there any hope of that being possible to implement? I assume that info does exist at runtime but I haven't found a way to extract it.

There is a way, but since even non-random-access collections can provide a faster `index(_:offsetBy:)`, I don't think we'd want to go that route…

PR #24, opened 2020-10-14
Add Stride for Collection

ollieatkinson:
Many thanks to @timvermeulen for continued help!

# Stride

A type that steps over a collection’s elements by the specified amount.

This is available through the `striding(by:)` method on any `Collection`.

```swift
(0...10).striding(by: 2) // == [0, 2, 4, 6, 8, 10]
```

If the stride is larger than the collection count, the resulting wrapper only contains the 
first element.

The stride amount must be a positive value.

## Detailed Design

The `striding(by:)` method is declared as a `Collection` extension, and returns a 
`Stride` type:

```swift
extension Collection {
  public func striding(by step: Int) -> Stride<Self>
}
```

A custom `Index` type is defined so that it's not possible to get confused when trying 
to access an index of the stride collection.

```swift
[0, 1, 2, 3, 4].striding(by: 2)[1] // == 1
[0, 1, 2, 3, 4].striding(by: 2).map { $0 }[1] // == 2
```

A careful thought was given to the composition of these strides by giving a custom 
implementation to `index(_:offsetBy:limitedBy)` which multiplies the offset by the 
stride amount. 

```swift
base.index(i.base, offsetBy: distance * stride, limitedBy: base.endIndex)
```

The following two lines of code are equivalent, including performance:

```swift
(0...10).striding(by: 6)
(0...10).striding(by: 2).stride(by: 3)
```

### Complexity

The call to `striding(by: k)` is always O(_1_) and access to the next value in the stride 
is O(_1_) if the collection conforms to `RandomAccessCollection`, otherwise O(_k_).

[rust has `Strided`](https://docs.rs/strided/0.2.9/strided/) available in a crate. 
[c++ has std::slice::stride](http://www.cplusplus.com/reference/valarray/slice/stride/)

The semantics of `striding` described in this documentation are equivalent.

---

### Checklist

### DIFF
--- a/Guides/Stride.md
+++ b/Guides/Stride.md
@@ -1,0 +1,63 @@
+# Stride
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Stride.swift) | 
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/StrideTests.swift)]
+
+A type that steps over a collection’s elements by the specified amount.
+
+This is available through the `striding(by:)` method on any `Collection`.
+
+```swift
+(0...10).striding(by: 2) // == [0, 2, 4, 6, 8, 10]
+```
+
+If the stride is larger than the collection count, the resulting wrapper only contains the 
+first element.
+
+The stride amount must be a positive value.
+
+## Detailed Design
+
+The `striding(by:)` method is declared as a `Collection` extension, and returns a 
+`Stride` type:
+
+```swift
+extension Collection {
+  public func striding(by step: Int) -> Stride<Self>
+}
+```
+
+A custom `Index` type is defined so that it's not possible to get confused when trying 
+to access an index of the stride collection.
+
+```swift
+[0, 1, 2, 3, 4].striding(by: 2)[1] // == 1
+[0, 1, 2, 3, 4].striding(by: 2).map { $0 }[1] // == 2
+```
+
+A careful thought was given to the composition of these strides by giving a custom 
+implementation to `index(_:offsetBy:limitedBy)` which multiplies the offset by the 
+stride amount. 
+
+```swift
+base.index(i.base, offsetBy: distance * stride, limitedBy: base.endIndex)
+```
+
+The following two lines of code are equivalent, including performance:
+
+```swift
+(0...10).striding(by: 6)
+(0...10).striding(by: 2).stride(by: 3)
+```
+
+### Complexity
+
+The call to `striding(by: k)` is always O(_1_) and access to the next value in the stride 
+is O(_1_) if the collection conforms to `RandomAccessCollection`, otherwise O(_k_).
+
+### Comparison with other languages
+
+[rust has `Strided`](https://docs.rs/strided/0.2.9/strided/) available in a crate. 
+[c++ has std::slice::stride](http://www.cplusplus.com/reference/valarray/slice/stride/)
+
+The semantics of `striding` described in this documentation are equivalent.
--- a/Guides/Stride.md
+++ b/Guides/Stride.md
@@ -5,7 +5,7 @@
 
 A type that steps over a collection’s elements by the specified amount.
 
-This is available through the `striding(by:)` method on any `Collection`.
+This is available through the `striding(by:)` method on any `Sequence`.
 
 ```swift
 (0...10).striding(by: 2) // == [0, 2, 4, 6, 8, 10]
@@ -18,11 +18,11 @@ The stride amount must be a positive value.
 
 ## Detailed Design
 
-The `striding(by:)` method is declared as a `Collection` extension, and returns a 
+The `striding(by:)` method is declared as a `Sequence` extension, and returns a 
 `Stride` type:
 
 ```swift
-extension Collection {
+extension Sequence {
   public func striding(by step: Int) -> Stride<Self>
 }
 ```
--- a/Guides/Stride.md
+++ b/Guides/Stride.md
@@ -3,7 +3,7 @@
 [[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Stride.swift) | 
  [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/StrideTests.swift)]
 
-A type that steps over a collection’s elements by the specified amount.
+A type that steps over sequence elements by the specified amount.
 
 This is available through the `striding(by:)` method on any `Sequence`.
 
@@ -11,7 +11,7 @@ This is available through the `striding(by:)` method on any `Sequence`.
 (0...10).striding(by: 2) // == [0, 2, 4, 6, 8, 10]
 ```
 
-If the stride is larger than the collection count, the resulting wrapper only contains the 
+If the stride is larger than the count, the resulting wrapper only contains the 
 first element.
 
 The stride amount must be a positive value.
--- a/Sources/Algorithms/Stride.swift
+++ b/Sources/Algorithms/Stride.swift
@@ -84,7 +84,7 @@ extension Stride: Collection {
   
   public func index(after i: Index) -> Index {
     precondition(i.base < base.endIndex, "Advancing past end index")
-    return index(i, offsetBy: 1, limitedBy: endIndex) ?? endIndex
+    return index(i, offsetBy: 1)
   }
 
   public func index(
--- a/Sources/Algorithms/Stride.swift
+++ b/Sources/Algorithms/Stride.swift
@@ -1,0 +1,161 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// striding(by:)
+//===----------------------------------------------------------------------===//
+
+extension Collection {
+  /// Returns a collection stepping through the elements every `step` starting
+  /// at the first value. Any remainders of the stride will be trimmed.
+  ///
+  ///     (0...10).striding(by: 2) // == [0, 2, 4, 6, 8, 10]
+  ///     (0...10).striding(by: 3) // == [0, 3, 6, 9]
+  ///
+  /// - Complexity: O(1). Access to successive values is O(1) if the
+  /// collection conforms to `RandomAccessCollection`; otherwise,
+  /// O(_k_), where _k_ is the striding `step`.
+  ///
+  /// - Parameter step: The amount to step with each iteration.
+  /// - Returns: Returns a collection stepping through the elements by the
+  /// specified amount.
+  public func striding(by step: Int) -> Stride<Self> {
+    Stride(base: self, stride: step)
+  }
+}
+
+public struct Stride<Base: Collection> {
+  
+  public let base: Base
+  public let stride: Int
+  
+  init(base: Base, stride: Int) {
+    precondition(stride > 0, "striding must be greater than zero")
+    self.base = base
+    self.stride = stride
+  }
+}
+
+extension Stride {
+  public func striding(by step: Int) -> Self {
+    Stride(base: base, stride: stride * step)
+  }
+}
+
+extension Stride: Collection {
+  
+  public struct Index: Comparable {
+    
+    let base: Base.Index
+    
+    init(_ base: Base.Index) {
+      self.base = base
+    }
+    
+    public static func < (lhs: Index, rhs: Index) -> Bool {
+      lhs.base < rhs.base
+    }
+  }
+  
+  public var startIndex: Index {
+    Index(base.startIndex)
+  }
+  
+  public var endIndex: Index {
+    Index(base.endIndex)
+  }
+  
+  public subscript(i: Index) -> Base.Element {
+    base[i.base]
+  }
+  
+  public func index(after i: Index) -> Index {
+    precondition(i.base < base.endIndex, "Advancing past end index")
+    return base.index(i.base, offsetBy: stride, limitedBy: base.endIndex)
+      .map(Index.init) ?? endIndex
+  }
+  
+  public func index(
+    _ i: Index,
+    offsetBy distance: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    base.index(i.base, offsetBy: distance * stride, limitedBy: limit.base)
+      .map(Index.init)
+  }
+
+  public var count: Int {
+    let limit = base.count - 1
+    return limit / stride + (limit < 0 ? 0 : 1)
+  }
+  
+  public func distance(from start: Index, to end: Index) -> Int {
+    let distance = base.distance(from: start.base, to: end.base)
+    return distance / stride + (distance % stride > 0 ? 1 : 0)
+  }
+  
+  public func index(_ i: Index, offsetBy distance: Int) -> Index {
+    precondition(distance <= 0 || i.base < base.endIndex, "Advancing past end index")
+    precondition(distance >= 0 || i.base > base.startIndex, "Incrementing past start index")
+    return Index(base.index(i.base, offsetBy: distance * stride))
+  }
+}
+
+extension Stride: BidirectionalCollection
+  where Base: RandomAccessCollection {
+  
+  public func index(before i: Index) -> Index {
+    precondition(i.base > base.startIndex, "Incrementing past start index")
+    if i == endIndex {
+      let count = base.count
+      precondition(count > 0, "Can't move before the starting index")
+      return Index(
+        base.index(base.endIndex, offsetBy: -((count - 1) % stride + 1))
+      )
+    } else {
+      guard let step = base.index(
+        i.base,
+        offsetBy: -stride,
+        limitedBy: startIndex.base
+      ) else {
+        fatalError("Incrementing past start index")
+      }
+      return Index(step)
+    }
+  }
+}
+
+extension Stride: RandomAccessCollection
+  where Base: RandomAccessCollection {}
+
+extension Stride: Equatable
+  where Base.Element: Equatable {
+  
+  public static func == (lhs: Stride, rhs: Stride) -> Bool {
+    lhs.elementsEqual(rhs, by: ==)
+  }
+  
+}
+
+extension Stride: Hashable
+  where Base.Element: Hashable {
+  
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(stride)
+    for element in self {
+      hasher.combine(element)
+    }
+  }
+  
+}
+
+extension Stride.Index: Hashable
+  where Base.Index: Hashable {}
--- a/Sources/Algorithms/Stride.swift
+++ b/Sources/Algorithms/Stride.swift
@@ -85,11 +85,6 @@ extension Stride: Collection where Base: Collection {
       self.base = base
     }
     
-    init?(_ base: Base.Index?) {
-      guard let base = base else { return nil }
-      self.base = base
-    }
-    
     public static func < (lhs: Index, rhs: Index) -> Bool {
       lhs.base < rhs.base
     }
@@ -122,7 +117,7 @@ extension Stride: Collection where Base: Collection {
     
     return n > 0
       ? offsetForward(i, offsetBy: n, limitedBy: limit)
-      : offsetBackward(i, offsetBy: n, limitedBy: limit)
+      : offsetBackward(i, offsetBy: -n, limitedBy: limit)
   }
   
   private func offsetForward(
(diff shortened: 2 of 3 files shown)


Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,102 @@
...
+  // TODO: Implement distance(from:to:) and index(_:offsetBy:)
+
+}
+
+extension Stride: BidirectionalCollection
+  where Base: BidirectionalCollection {
+  
+  public func index(before i: Index) -> Index {
+    base.index(i.base, offsetBy: -stride, limitedBy: startIndex.base)
+      .map(Index.init) ?? startIndex
```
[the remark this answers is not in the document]
ollieatkinson: Blast! I should have caught this, the original implementation for this pre-calculated the endIndex based on the stride `step`, but I did not want to allocate additional storage where it was not needed. Let put this back into draft and I will re-think how to calculate traversing backwards. Thanks @natecook1000 

```swift
self.endIndex = base.index(base.startIndex, offsetBy: (Swift.max(0, limit) / stride + (limit < 0 ? 0 : 1)) * stride + 1)
```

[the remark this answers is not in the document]
ollieatkinson: https://github.com/apple/swift-algorithms/pull/24/commits/18bd60de6974c6f105486c321343d117db7693e6#diff-c182736f044b7c26cf3b2d09697442eab53f068d6fe175a53b01aaf65b6528eaR98

I have also added some test cases to cover this scenario:
https://github.com/apple/swift-algorithms/pull/24/files#diff-c17f1930369137556a2c3a771fd1eddf48154598141bac6ea335ac7bfec9b170R60

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,112 @@
...
+
+extension Collection {
+  /// - Complexity: O(_1_) and access to the next value in the stride
+  /// is O(_1_) if the collection conforms to `RandomAccessCollection`,
+  /// otherwise O(_k_) where `k` is the stride `step`.
+  ///
+  /// - Parameter step: The amount to step with each iteration.
+  /// - Returns: Returns a collection stepping through the elements by the
+  /// specified amount.
+  public func stride(by step: Int) -> Stride<Self> {
```
xwu, line 24: I wonder if this would best be `striding` to fit the ed/ing rule for naming nonmutating members. That said, I also worry about ambiguity with the existing `StrideTo`/`StrideThrough` types: are there other terms that are often used for this operation?

ollieatkinson: I am in agreement that `striding` is probably a better fit for this name for the reasons you've mentioned. I'm a little unsure that renaming the type to something else would be a good idea as the notion of a stride is widely known and something which is obvious as soon as you see it. In one of my earlier implementations I had named the type `StridingCollection` which I would not be against being explicit about, I removed the suffix `Collection` to keep in step with the rest of the algorithms in this repository. It would be good to see what other feedback I get regarding the naming of this before I change it. Thanks for your thoughts @xwu

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,112 @@
...
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// striding(by:)
+//===----------------------------------------------------------------------===//
+
+extension Collection {
+  /// - Complexity: O(_1_) and access to the next value in the striding
+  /// is O(_1_) if the collection conforms to `RandomAccessCollection`,
+  /// otherwise O(_k_) where `k` is the striding `step`.
```
natecook1000, line 19:
```suggestion
  /// - Complexity: O(1). Access to successive values is O(1) if the
  /// collection conforms to `RandomAccessCollection`; otherwise,
  /// O(_k_), where _k_ is the striding `step`.
```

Also, the documentation should start with a single sentence describing what the method returns.

natecook1000 (changes requested):
This is looking good, @ollieatkinson — thanks for this addition!

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,112 @@
...
+}
+
+extension Stride: RandomAccessCollection
+  where Base: RandomAccessCollection {}
+
+extension Stride: Equatable
+  where Base: Equatable {}
+
+extension Stride: Hashable
+  where Base: Hashable, Base.Index: Hashable {}
```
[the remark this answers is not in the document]
ollieatkinson: That is certainly interesting and something I would like to cover, I've given the implementation a go for `==` and `hash(into:)` in the latest commit. I would expect the collections to be equal iff their count is equal and all their elements are equal

Guides/Stride.md
```diff
@@ -0,0 +1,63 @@
...
+
+```swift
+base.index(i.base, offsetBy: distance * stride, limitedBy: base.endIndex)
+```
+
+The following two lines of code are equivalent, including performance:
+
+```swift
+(0...10).striding(by: 6)
+(0...10).striding(by: 2).stride(by: 3)
```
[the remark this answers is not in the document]
ollieatkinson: I like it!

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,112 @@
...
+  public func index(
+    _ i: Index,
+    offsetBy distance: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    base.index(i.base, offsetBy: distance * stride, limitedBy: limit.base)
+      .map(Index.init)
+  }
+  
+  // TODO: Implement distance(from:to:) and index(_:offsetBy:)
```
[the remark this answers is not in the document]
ollieatkinson: They are quite simple functions to implement, so I've addressed all 3 in the latest commit:

```swift
  public var count: Int {
    (base.count - 1) / stride + 1
  }
  
  public func distance(from start: Index, to end: Index) -> Int {
    base.distance(from: start.base, to: end.base) / stride
  }
  
  public func index(_ i: Index, offsetBy distance: Int) -> Index {
    Index(base.index(i.base, offsetBy: distance * stride))
  }
```

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,160 @@
...
+    base.index(i.base, offsetBy: distance * stride, limitedBy: limit.base)
+      .map(Index.init)
+  }
+  
+  public var count: Int {
+    (base.count - 1) / stride + 1
+  }
+  
+  public func distance(from start: Index, to end: Index) -> Int {
+    base.distance(from: start.base, to: end.base) / stride
```
natecook1000, line 100: This can have an off by one error when `end == endIndex` and `base.count` isn’t a multiple of `stride`.

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,160 @@
...
+  }
+  
+  public func index(_ i: Index, offsetBy distance: Int) -> Index {
+    Index(base.index(i.base, offsetBy: distance * stride))
```
natecook1000, line 104: Similar issue — this can overrun `base.endIndex` in a way that will trap in some collections when it should just return `endIndex`. I think this will have the problem:

```swift
let a = 1...5
let b = a.striding(by: 3) // equivalent to [1, 4]
let i = b.index(b.startIndex, offsetBy: 2)
```

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,160 @@
...
+    _ i: Index,
+    offsetBy distance: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    base.index(i.base, offsetBy: distance * stride, limitedBy: limit.base)
+      .map(Index.init)
+  }
+  
+  public var count: Int {
+    (base.count - 1) / stride + 1
```
natecook1000, line 96: I don’t think this works when the collection is empty — `(0 - 1) / stride + 1` is going to be `1` when `stride > 1`.

Tests/SwiftAlgorithmsTests/StrideTests.swift
```diff
@@ -0,0 +1,105 @@
...
+  
+  func testCount() {
+    let a = (0...10)
+    XCTAssertEqual(a.striding(by: 1).count, (0...10).count)
+    XCTAssertEqual(a.striding(by: 2).count, [0, 2, 4, 6, 8, 10].count)
+    XCTAssertEqual(a.striding(by: 3).count, [0, 3, 6, 9].count)
+    XCTAssertEqual(a.striding(by: 4).count, [0, 4, 8].count)
+    XCTAssertEqual(a.striding(by: 5).count, [0, 5, 10].count)
+    XCTAssertEqual(a.striding(by: 10).count, [0, 10].count)
+    XCTAssertEqual(a.striding(by: 11).count, [0].count)
```
natecook1000, line 86: Ah — yes:
```suggestion
    XCTAssertEqual(a.striding(by: 11).count, 0)
```

Tests/SwiftAlgorithmsTests/StrideTests.swift
```diff
@@ -0,0 +1,105 @@
...
+  }
+  
+  func testDistance() {
+    let a = (0...100).striding(by: 12)
+    var i = a.startIndex
+    a.formIndex(&i, offsetBy: 3)
+    XCTAssertEqual(a.distance(from: a.startIndex, to: i), 3)
+    XCTAssertEqual(a[i], 36)
+    XCTAssertEqual(a.distance(from: a.startIndex, to: a.endIndex), a.count - 1)
```
natecook1000, line 95: Here too:
```suggestion
    XCTAssertEqual(a.distance(from: a.startIndex, to: a.endIndex), a.count)
```

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,160 @@
...
+}
+
+extension Stride: RandomAccessCollection
+  where Base: RandomAccessCollection {}
+
+extension Stride: Equatable
+  where Base.Element: Equatable {
+  
+  public static func == (lhs: Stride, rhs: Stride) -> Bool {
+    guard lhs.count == rhs.count else { return false }
```
natecook1000, line 140: You can leave this check out — `elementsEqual` will fail if they have different lengths.

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,160 @@
...
+    return lhs.elementsEqual(rhs, by: ==)
+  }
+  
+}
+
+extension Stride: Hashable
+  where Base.Element: Hashable {
+  
+  public func hash(into hasher: inout Hasher) {
+    hasher.combine(count)
```
natecook1000, line 150: You can leave the count out here, too.

natecook1000 (commented):
@ollieatkinson A few more notes for you!

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,160 @@
...
+  /// - Returns: Returns a collection stepping through the elements by the
+  /// specified amount.
+  public func striding(by step: Int) -> Stride<Self> {
+    Stride(base: self, stride: step)
+  }
+}
+
+public struct Stride<Base: Collection> {
+  
+  public let base: Base
```
xwu, line 37: To help simplify the issue, do we need to expose `base` at all?

Sources/Algorithms/Stride.swift
ollieatkinson: I was a little unsure about what to do here, when using the standard collection API for `offsetBy` it trapped as you say so I mimic'd that behaviour.

Is it better to phrase such that it returns `endIndex` like such:
```swift
  public func index(_ i: Index, offsetBy distance: Int) -> Index {
    precondition(i.base < base.endIndex, "Advancing past end index")
    let limit = distance > 0 ? endIndex : startIndex
    return index(i, offsetBy: distance, limitedBy: limit) ?? limit
  }
```

Sources/Algorithms/Stride.swift
ollieatkinson: I was also unsure of the value of being able to refer to `base` in this instance, but since all of the other collection algorithms do I thought it would be worth keeping consistent. Was there a specific issue you think that it might simplify?

xwu: Sorry, the specific issue is the question of equality, which @natecook1000 mentioned above.

Two values are equal when they behave the same way for all salient operations. For this purpose, we say that iteration order isn’t salient for sets but is salient for arrays, that the sign of zero isn’t salient for floating-point numbers, etc.

It’s a judgment call about what is salient here, and reasonable people can disagree whether differing `base` elements would matter; by contrast, if `base` isn’t recoverable, then there is no ambiguity at all.

ollieatkinson: I like what you're saying here and I'd be up for dropping the access modifier from the property - I cannot fathom a situation where you would want to use it outside of the scope of the collection itself.

natecook1000 (commented):
I think just these last couple of notes, then this will be ready 👍

ollieatkinson:
I've rebased against the latest `main` and included some test cases using the new `validateIndexTraversals` TestUtilities API. Before merging, it would be good to hear your thoughts on the points that xwu raised [here](https://github.com/apple/swift-algorithms/pull/24#discussion_r509542587)

Sources/Algorithms/Stride.swift
natecook1000: Agreed on this — `base` can be internal, which removes any controversy here. (That should probably be the case for the rest of the collection wrappers as well.)

ollieatkinson: Thanks both for your thoughts, [I've removed it](https://github.com/apple/swift-algorithms/pull/24/commits/dcdef508c0abc1839008ae14b7d6d6e960164735) from base here and we can follow up in another P/R for the rest of the collections

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,161 @@
...
+  
+  public func distance(from start: Index, to end: Index) -> Int {
+    let distance = base.distance(from: start.base, to: end.base)
+    return distance / stride + (distance % stride > 0 ? 1 : 0)
+  }
+  
+  public func index(_ i: Index, offsetBy distance: Int) -> Index {
+    precondition(i.base < base.endIndex, "Advancing past end index")
+    let limit = distance > 0 ? endIndex : startIndex
+    return index(i, offsetBy: distance, limitedBy: limit) ?? limit
```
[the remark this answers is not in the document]
ollieatkinson: The precondition has been added in the case of `nil` https://github.com/apple/swift-algorithms/pull/24/commits/6df6133ab7400b2b5a8c6fe40a94a4f74178cc43#diff-c182736f044b7c26cf3b2d09697442eab53f068d6fe175a53b01aaf65b6528eaR132

Sources/Algorithms/Stride.swift
timvermeulen: >by contrast, if `base` isn’t recoverable, then there is no ambiguity at all.

I don't see how this matters in case the user knows what `base` is, even if `Stride` doesn't expose it. Couldn't there still be controversy about whether `(0...10).striding(by: 4)` should equal `(0...11).striding(by: 4)`?

Tests/SwiftAlgorithmsTests/StrideTests.swift
```diff
@@ -0,0 +1,127 @@
...
+  }
+  
+  func testIndexTraversals() {
+    let zero_to_one_hundered_range = 0...100
+    validateIndexTraversals(
+      zero_to_one_hundered_range.striding(by: 10),
+      zero_to_one_hundered_range.striding(by: 11),
+      zero_to_one_hundered_range.striding(by: 101)
+    )
+    let zero_to_one_hundered_array = zero_to_one_hundered_range.map{ $0 }
```
timvermeulen, line 98:
```suggestion
    let zero_to_one_hundered_array = Array(zero_to_one_hundered_range)
```

Tests/SwiftAlgorithmsTests/StrideTests.swift
```diff
@@ -0,0 +1,127 @@
...
+      string.striding(by: 2),
+      string.striding(by: 10)
+    )
+  }
+
+  
+  func testOffsetBy() {
+    let a = (0...100).striding(by: 22)
+    let b = [0, 22, 44, 66, 88]
+    for i in 0..<a.count  {
```
timvermeulen, line 116:
```suggestion
    for i in 0..<a.count {
```

Tests/SwiftAlgorithmsTests/StrideTests.swift
```diff
@@ -0,0 +1,127 @@
...
+      zero_to_one_hundered_array.striding(by: 11),
+      zero_to_one_hundered_array.striding(by: 101)
+    )
+    let string = "swift rocks".map(String.init)
+    validateIndexTraversals(
+      string.striding(by: 1),
+      string.striding(by: 2),
+      string.striding(by: 10)
+    )
+  }
```
timvermeulen, line 110: We should probably also have a traversal test with an empty base collection.

timvermeulen:
`Stride` probably shouldn't require that `Base: Collection`, but instead implement only `Sequence` when `Base: Sequence`.

ollieatkinson:
@timvermeulen 
> Stride probably shouldn't require that Base: Collection, but instead implement only Sequence when Base: Sequence.

For this, are you implying the following:

```swift
public struct Stride<Base: Sequence>
```

```swift
extension Stride: Sequence {
  public struct Iterator: IteratorProtocol {
    var iterator: Base.Iterator
    let stride: Int
    public mutating func next() -> Base.Element? {
      defer {
        for _ in 0..<stride-1 where iterator.next() != nil {}
      }
      return iterator.next()
    }
  }
  
  public func makeIterator() -> Stride<Base>.Iterator {
    Iterator(iterator: base.makeIterator(), stride: stride)
  }
}
```

And that conformance to `Collection` is only applicable when the base collection is also `Collection`?

```swift
extension Stride: Collection where Base: Collection
```

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,224 @@
...
+    var iterator: Base.Iterator
+    let stride: Int
+    
+    public mutating func next() -> Base.Element? {
+      defer {
+        for _ in 0..<(stride - 1) {
+          guard iterator.next() != nil else { break }
+        }
+      }
+      return iterator.next()
```
timvermeulen, line 66: This always immediately computes the `stride - 1` elements after the one being returned. Would it be better to compute them only when the element after them is being returned, instead?

ollieatkinson: Yes it does. I had another implementation which deferred the compute, it does add more storage onto the iterator. With the worst case being an extra O(k-1), if `k` is large and unbounded it may cause problems so I think I will look again toward this approach

```swift
extension Stride: Sequence {
  
  public struct Iterator: IteratorProtocol {

    var iterator: Base.Iterator
    let stride: Int
    var striding: Bool = false
    
    public mutating func next() -> Base.Element? {
      
      guard striding else {
        striding = true
        return iterator.next()
      }
      
      for _ in 0..<(stride - 1) {
        guard iterator.next() != nil else { break }
      }
      return iterator.next()
    }
  }
  
  public func makeIterator() -> Stride<Base>.Iterator {
    Iterator(iterator: base.makeIterator(), stride: stride)
  }
}
```

timvermeulen: This looks reasonable, iterators are almost always temporary so it should be nog big deal to add a `Bool` to it.

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,224 @@
...
+    _ i: Index,
+    offsetBy n: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    guard n != 0 else { return i }
+    guard limit != i else { return nil }
+    
+    return n > 0
+      ? offsetForward(i, offsetBy: n, limitedBy: limit)
+      : offsetBackward(i, offsetBy: n, limitedBy: limit)
```
timvermeulen, line 125: I would suggest passing `-n` here to have both `offsetForward` and `offsetBackward` be given a positive offset, similar to how `Chain2` currently does it.

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,224 @@
...
+    let base: Base.Index
+    
+    init(_ base: Base.Index) {
+      self.base = base
+    }
+    
+    init?(_ base: Base.Index?) {
+      guard let base = base else { return nil }
+      self.base = base
+    }
```
timvermeulen, line 91: I think you're only using this once at the moment, let's get rid of it and just call `.map(Index.init)` at the point of use.

ollieatkinson:
@natecook1000 Is there any more work required on this for it to be merged?

Sources/Algorithms/Stride.swift
```diff
@@ -0,0 +1,220 @@
...
+  }
+  
+  private func offsetBackward(
+    _ i: Index,
+    offsetBy n: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    let distance = i == endIndex
+      ? -((base.count - 1) % stride + 1) + (n - 1) * -stride
```
jshier, line 158: FYI, Xcode 12.2 fails to compile this line due to an "expression too complex" error. Making the `1` values `Int(1)` seems to resolve the issue.

ollieatkinson: Thanks for raising this, I will take a look tomorrow AM - I'm surprised that it's too complex given these are all integers. 
I can probably annotate the `distance` property `: Int` to the same effect.

natecook1000 (approved):

natecook1000:
Thanks, @ollieatkinson — this is ready to go!

PR #26, opened 2020-10-15
Fix `Product2`'s indexing logic

timvermeulen:
This PR changes the way `Product2.endIndex` is represented: instead of `(base1.endIndex, base2.endIndex)`, it is now `(base1.endIndex, base2.startIndex)`.

The latter turns out to be a more natural way of representing the "past the end" index because it is exactly the index after the last valid index `(lastValidIndexOfBase1, lastValidIndexOfBase2)` according to the existing mechanism of advancing any other index of `Product2`. As a result, the methods `index(before:)`, `index(after:)`, and `distance(from:to:)` don't need any custom logic to deal with `endIndex`.

`endIndex` wasn't handled correctly everywhere so this also ends up fixing bugs in `index(before:)` and `distance(from:to:)`.

### Checklist

### DIFF
--- a/Sources/Algorithms/Product.swift
+++ b/Sources/Algorithms/Product.swift
@@ -120,7 +120,7 @@ extension Product2: Collection where Base1: Collection {
     }
     
     return base2[start.i2...].count + base2[..<end.i2].count
-      + base2.count * (base1.distance(from: start.i1, to: end.i1))
+      + base2.count * (base1.distance(from: start.i1, to: end.i1) - 1)
   }
 
   public subscript(position: Index) -> (Base1.Element, Base2.Element) {
--- a/Sources/Algorithms/Product.swift
+++ b/Sources/Algorithms/Product.swift
@@ -135,12 +135,12 @@ extension Product2: BidirectionalCollection
     precondition(i != startIndex,
                  "Can't move before startIndex")
     if i.i2 == base2.startIndex {
+      return Index(
+        i1: base1.index(before: i.i1),
+        i2: base2.index(before: base2.endIndex))
+    } else {
       return Index(i1: i.i1, i2: base2.index(before: i.i2))
     }
-    
-    return Index(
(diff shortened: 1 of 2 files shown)


Sources/Algorithms/Product.swift
```diff
@@ -91,27 +91,22 @@ extension Product2: Collection where Base1: Collection {
...
-      ? endIndex
-      : Index(i1: base1.startIndex, i2: base2.startIndex)
+    Index(
+      i1: base2.isEmpty ? base1.endIndex : base1.startIndex,
+      i2: base2.startIndex)
   }
   
   public var endIndex: Index {
-    Index(i1: base1.endIndex, i2: base2.endIndex)
+    // this representation makes index calculations simpler
```
kylemacomber, line 100: I'm not sure we need this comment to make it into the final patch

kylemacomber (commented):
This looks like a great simplification. Nice work @timvermeulen!

Sources/Algorithms/Product.swift
timvermeulen: Thanks for the feedback, the only reason for putting it there was to indicate to someone reading the code that this is not a typo – any idea how else we could convey this?

kylemacomber: I'll defer to @natecook1000 on whether we should keep it, I guess I generally don't assume there are typos in the code, and if I'm surprised I'll read the surrounding code closer.

That being said, I can understand wanting to draw attention to i2 being base.startIndex, to prevent the reader from glossing over it. For example:

```
// `base2.startIndex` simplifies index calculations.
```

timvermeulen: Thanks, that's indeed what I was going for. I'll change it to that for now and then leave it up to Nate whether to remove it completely.

natecook1000:
Thanks, @timvermeulen! :shipit:

PR #28, opened 2020-10-16
Use Swift Numerics for elementary functions

stephentyrone:
This drops the canImport darwin/glibc/etc dance, which is ugly, and lets us use `root` instead of `exp(log(x)/k)`, which is more accurate in some cases (but mainly is just nicer to read).
    
The downside is that this introduces a dependency for Algorithms, where previously it had none. I think that Numerics is an OK thing to depend on (especially for these functions, which are available from 0.0.1), but it _is_ a _massive increase_ in the number of dependencies that the package has, and we should give that some thought.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,11 +20,14 @@ let package = Package(
             targets: ["Algorithms"]),
     ],
     dependencies: [
+        .package(url: "https://github.com/apple/swift-numerics", from: "0.0.1"),
     ],
     targets: [
         .target(
             name: "Algorithms",
-            dependencies: []),
+            dependencies: [
+              .product(name: "RealModule", package: "swift-numerics"),
+            ]),
         .testTarget(
             name: "SwiftAlgorithmsTests",
             dependencies: ["Algorithms"]),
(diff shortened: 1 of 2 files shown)


natecook1000:
I'm definitely reticent to add this due to the, as you say, massive increase in dependencies. It seems like there's a trade-off at some point — if `Algorithms` were to use more of the higher-level functionality, or the generic capability that Swift Numerics provides, it would make sense to bring in the dependency, but at this point it isn't worth the addition. What do you think?

stephentyrone:
I think it makes sense because:
(a) We want people to use SN for the elementary functions--it's the "blessed" package for those APIs. If Numerics needed Cycle or whatever, I would expect to pull in Algorithms for it.
(b) In this particular case, I expect that Algorithms will want other pieces of Numerics in the future (especially stuff around RNGs), so I believe that you'll end up with this dependency sooner or later anyway.

natecook1000:
> (c) If we ever want to get the same results across platforms for a given PRNG (which I expect we will), the only path to that is via Numerics. We could choose to provide portable log and root operations there, but the Darwin/Glibc/etc operations will never give the same results everywhere.

Sold on this point!

PR #30, opened 2020-10-20
Fix stablePartition(subrange:by:) using subrange count instead of entire collection count.

pmtao:
Function `stablePartition(subrange:by:)` in `Partition` should not using the collection count which will cause bound error, but using subrange count.

### Checklist


kylemacomber:
Good catch!

natecook1000:
Thanks for this fix, @pmtao! :clap::clap::clap:

PR #31, opened 2020-10-21
Add Sequence.firstNonNil(_:)

### LINKED ISSUE #21: A `first` operator that returns the first non-nil transformed value
I found myself often having to find the first element in an array that passes an optional initialiser.
If I wanted to stick to a purely functional paradigm (which I do, I'm a zealot) I need to do some pretty not-ideal transforms to allow this behaviour without defining a new function.
```
let value = ["Hello", "10"].lazy.map(Int.init).first { $0 != nil }.flatMap { $0 }
```
This could be trivialised with the following (naive) implementation.
```
public extension Sequence {
    /// Returns the first element in `self` that `transform` maps to a `.some`.
    func first<Result>(of transform: (Element) throws -> Result?) rethrows -> Result? {
        for value in self {
            if let value = try transform(value) {
                return value
            }
        }
        return nil
    }
}
```

Qata:
Resolves #21 

### Description
`firstNonNil(_:)` returns the first value that is resolved by the passed transform to a non `nil` (`.some`) value.

### Detailed Design

```swift
extension Sequence {
    /// Returns the first element in `self` that `transform` maps to a `.some`.
    func firstNonNil<Result>(_ transform: (Element) throws -> Result?) rethrows -> Result?
}
```

### Documentation Plan
Added documentation comments to the new function.

### Test Plan
Five tests have been added.

### Source Impact
None

### Checklist


Sources/Algorithms/First.swift
```diff
@@ -0,0 +1,25 @@
...
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// firstNonNil(of:)
+//===----------------------------------------------------------------------===//
+public extension Sequence {
+    /// Returns the first element in `self` that `transform` maps to a `.some`.
+    func firstNonNil<Result>(of transform: (Element) throws -> Result?) rethrows -> Result? {
```
ktoso, line 17: The function name sounds good (as bikeshedded in the ticket), but are we sure about the `of`? Reads a bit weird.
It's more like a `where` that happens to be implemented with `some/none` rather than `true/false` but it's still a form of predicate closer to where rather than "of". 

wdyt?

ktoso (commented):
Looks okey :) Should probably also gain a `Guides/` doc?

Minor: Maybe consider keeping the ticket open and adding the usual Resolves https://github.com/apple/swift-algorithms/issues/21 here rather than closing ticket immediately?

Sources/Algorithms/First.swift
natecook1000: Agreed. I think we can drop the label — to me it reads better when it's more similar to `map`, e.g.: `guard let number = inputs.firstNonNil(Int.init) else { ... }`

Qata: This feature felt minimal and self-explanatory enough to not be worth a guides doc, but I can add one.

Qata:
> Looks okey :) Should probably also gain a `Guides/` doc?
> 
> Minor: Maybe consider keeping the ticket open and adding the usual Resolves #21 here rather than closing ticket immediately?

This feature felt minimal and self-explanatory enough to not be worth a guides doc, but I'll add one!

Guides/First.md
```diff
@@ -0,0 +1,45 @@
+# First
```
xwu, line 1: Since this API isn't `first`, but `firstNonNil`, can we align the naming of documents and headings accordingly?

Qata: My thinking that it was a general category of functions for which others might be added. I'll amend it.

Guides/FirstNonNil.md
```diff
@@ -0,0 +1,43 @@
+# FirstNonNil
```
xwu, line 1:
```suggestion
# First Non-Nil
```

Sources/Algorithms/FirstNonNil.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// firstNonNil(_:)
+//===----------------------------------------------------------------------===//
+
+public extension Sequence {
+    /// Returns the first element in `self` that `transform` maps to a `.some`.
+    func firstNonNil<Result>(_ transform: (Element) throws -> Result?) rethrows -> Result? {
```
xwu, line 18: Formatting: to be consistent with the rest of the project, please wrap to 80 columns. Also, use 2 spaces instead of 4 for indenting.

```suggestion
  func firstNonNil<Result>(
    _ transform: (Element) throws -> Result?
  ) rethrows -> Result? {
```

xwu (commented):
It looks like we’ve settled on a name of `firstNonNil(_:)`. Can you update the pull request title and message? They currently say that we’re adding an API called `firstNonNil(of:)`.

Guides/FirstNonNil.md
```diff
@@ -0,0 +1,43 @@
...
+let value = ["A", "B", "10"].firstNonNil { Int($0) }
+// value == .some(10)
+// 
+let noValue = ["A", "B", "C"].firstNonNil { Int($0) }
+// noValue == .none
+```
+
+
+This method is analogous to `first(where:)` in how it only consumes values until
+a `.some` is found, unlike using lazy operators, which will load any sequence into a collection
```
xwu, line 20: Could you be a little more explicit here about why this method is useful? I think @timvermeulen explained it very well in the corresponding issue, and if you get his permission, that might be a helpful piece of information for users.

Guides/FirstNonNil.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+### Naming
+
+This method’s name was selected for its comprehensibility. 
+
+### Comparison with other languages
+
+**Scala**: Scala provides a `collectFirst` function that finds the first element
+in a collection for which a partial function is defined.
```
xwu, line 43: There were a few other precedents mentioned in the corresponding issue. Could you please include them here? Ideally, we’d look around a bit more and really study the landscape here so that we can be sure we’ve picked a good name.

Guides/FirstNonNil.md
```diff
@@ -0,0 +1,43 @@
...
+public extension Sequence {
+    func firstNonNil<Result>(_ transform: (Element) throws -> Result?)
+    rethrows -> Result? 
+}
+
+```
+
+### Naming
+
+This method’s name was selected for its comprehensibility.
```
xwu, line 38: I don’t think this sentence adds much to the reader. If this section isn’t mandatory, then it can be deleted. If it _is_ mandatory, then you’ll need to explain _why_ this name has good “comprehensibility.” Perhaps @natecook1000 can help you to explain since he made the choice.

natecook1000: It would be great to capture the other names that came up in the issue discussion here. Why  was `firstNonNil` chosen over those?

Sources/Algorithms/FirstNonNil.swift
```diff
@@ -0,0 +1,28 @@
...
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// firstNonNil(_:)
+//===----------------------------------------------------------------------===//
+
+public extension Sequence {
+    /// Returns the first element in `self` that `transform` maps to a `.some`.
```
natecook1000, line 17: This method doesn't return the first element, it returns the mapped result. Can you update this wording and include an example demonstrating the usage?

Sources/Algorithms/FirstNonNil.swift
```diff
@@ -0,0 +1,28 @@
...
+  func firstNonNil<Result>(
+    _ transform: (Element) throws -> Result?
+  ) rethrows -> Result? {
+        for value in self {
+            if let value = try transform(value) {
+                return value
+            }
+        }
+        return nil
+    }
```
natecook1000, line 27: Nit: For source files we use 2-space indents.

timvermeulen:
Hey @Qata, this is to let you know we'll soon be landing this addition and making the minor changes necessary in order to ship this in the next release. 🚀 Thank you for your work on this, and feel free to make any changes you still want to make in the meantime.

timvermeulen:
@swift-ci please test

PR #32, opened 2020-10-22
Added documentation and where clause for API `uniqued()`

Roshankumar350:
Problem: 
    * A quick look up for the API named `uniqued()` with example  associated with it was missing.
    * API named `uniqued()` while enumerating each element was checking an `if` condition.

Solution:
    * API documentation added to address the same.
    *  `where` clause added for the same

### Checklist
- [ yes] I've added at least one test that validates that my change is working, if appropriate
- [ yes] I've followed the code style of the rest of the project
- [ yes] I've read the [Contribution Guidelines](../CONTRIBUTING.md)
- [ yes] I've updated the documentation if necessary

### DIFF
--- a/Sources/Algorithms/Unique.swift
+++ b/Sources/Algorithms/Unique.swift
@@ -25,15 +25,9 @@ extension Sequence where Element: Hashable {
   /// - Returns: An array with only the unique elements of this sequence.
   ///  .
   /// - Complexity: O(*n*), where *n* is the length of the sequence.
+  @inlinable
   public func uniqued() -> [Element] {
-    var seen: Set<Element> = []
-    var result: [Element] = []
-    for element in self {
-      if seen.insert(element).inserted {
-        result.append(element)
-      }
-    }
-    return result
+    uniqued(on: { $0 })
   }
 }
 
@@ -59,6 +53,7 @@ extension Sequence {
   ///   determined by the result of `projection` for each element.
   ///
   /// - Complexity: O(*n*), where *n* is the length of the sequence.
+  @inlinable
   public func uniqued<Subject: Hashable>(
     on projection: (Element) throws -> Subject
   ) rethrows -> [Element] {
--- a/Sources/Algorithms/Unique.swift
+++ b/Sources/Algorithms/Unique.swift
@@ -1,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// uniqued()
+//===----------------------------------------------------------------------===//
+
+extension Sequence where Element: Hashable {
+  /// Returns an array with only the unique elements of this sequence, in the
+  /// order of the first occurence of each unique element.
+  public func uniqued() -> [Element] {
+    var seen: Set<Element> = []
+    var result: [Element] = []
+    for element in self {
+      if seen.insert(element).inserted {
+        result.append(element)
+      }
+    }
+    return result
+  }
+}
+
+extension Sequence {
+  /// Returns an array with the unique elements of this sequence (as determined
+  /// by the given projection), in the order of the first occurence of each
+  /// unique element.
+  public func uniqued<Subject: Hashable>(
+    on projection: (Element) throws -> Subject
+  ) rethrows -> [Element] {
+    var seen: Set<Subject> = []
+    var result: [Element] = []
+    for element in self {
+      if seen.insert(try projection(element)).inserted {
+        result.append(element)
+      }
+    }
+    return result
+  }
+}
--- a/Sources/Algorithms/Unique.swift
+++ b/Sources/Algorithms/Unique.swift
@@ -15,7 +15,16 @@
 
 extension Sequence where Element: Hashable {
   /// Returns an array with only the unique elements of this sequence, in the
-  /// order of the first occurence of each unique element.
+  /// order of the first occurrence of each unique element.
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "dog", "cat"]
+  ///     let uniqued = animals.uniqued()
+  ///     print(uniqued)
+  ///     // Prints '["dog", "pig", "cat", "ox"]'
+  ///
+  /// - Returns: An array with only the unique elements of this sequence.
+  ///  .
+  /// - Complexity: O(*n*), where *n* is the length of the sequence.
   public func uniqued() -> [Element] {
     var seen: Set<Element> = []
     var result: [Element] = []
@@ -30,8 +39,26 @@ extension Sequence where Element: Hashable {
 
 extension Sequence {
   /// Returns an array with the unique elements of this sequence (as determined
-  /// by the given projection), in the order of the first occurence of each
+  /// by the given projection), in the order of the first occurrence of each
   /// unique element.
+  ///
+  /// This example finds the elements of the `animals` array with unique
+  /// first characters:
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "cow", "owl"]
+  ///     let uniqued = animals.uniqued(on: {$0.first})
+  ///     print(uniqued)
+  ///     // Prints '["dog", "pig", "cat", "ox"]'
+  ///
+  /// - Parameter projection: A closure that transforms an element into the
+  ///   value to use for uniqueness. If `projection` returns the same value
+  ///   for two different elements, the second element will be excluded
+  ///   from the resulting array.
+  ///
+  /// - Returns: An array with only the unique elements of this sequence, as
+  ///   determined by the result of `projection` for each element.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the sequence.
   public func uniqued<Subject: Hashable>(
     on projection: (Element) throws -> Subject
   ) rethrows -> [Element] {
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -1,0 +1,33 @@
+---
+name: 🐛 Bug Report
+about: Something isn't working as expected
+---
+
+<!--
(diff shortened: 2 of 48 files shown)


xwu:
Good catch on the typo!

>  I've followed the code style of the rest of the project

Note that the code style involves indentations with _two_ spaces, and also wrapping to 80 columns on each line.

Sources/Algorithms/Unique.swift
```diff
@@ -14,33 +14,57 @@
 //===----------------------------------------------------------------------===//
 
 extension Sequence where Element: Hashable {
-  /// Returns an array with only the unique elements of this sequence, in the
-  /// order of the first occurence of each unique element.
+  /// Returns an array with only  unique elements of this sequence, in the
+  /// order of the first occurrence of each unique element.
+  ///
+  /// In this example, `uniqued()` is used to include only names which are
+  /// unique
```
xwu, line 21: Thanks, the formatting is now mostly consistent. Make sure to include periods at the end of sentences, like here (also elsewhere.) There is still another closing brace that is incorrectly indented. There are also several places where you have two consecutive spaces in the comment or the code; please make sure to remove them. There are also places where there are missing spaces after punctuation: please make sure to add them.

A more general comment here: I don’t think it adds anything to describe the example here by saying that `uniqued` makes the array unique.

Sources/Algorithms/Unique.swift
```diff
@@ -14,33 +14,57 @@
...
+  /// by the given projection), in the order of the first occurrence of each
   /// unique element.
+  ///
+  /// In this example, `uniqued()` is used to include only names which are
+  /// unique
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "dog","cat"]
+  ///     let uniqued = animals.uniqued()
+  ///     print(uniqued)
+  ///     // Prints "["dog", "pig","cat", "ox"]"`
```
natecook1000, line 52: Could you switch this to an example that demonstrates the `uniqued(on:)` method? This example shows the parameterless method defined above.

Sources/Algorithms/Unique.swift
```diff
@@ -14,33 +14,57 @@
...
+  ///
+  /// - Parameter projection: A projecting closure. `projection` accepts an
+  ///   element of this sequence as its parameter and returns a projected
+  ///   value of the same  type.
+  ///
+  /// - Returns: Returns an array with only the unique elements of this sequence
```
natecook1000, line 58:
```suggestion
  /// - Returns: An array with only the unique elements of this sequence, as
  ///   determined by the result of `projection` for each element.
```

Sources/Algorithms/Unique.swift
```diff
@@ -14,33 +14,57 @@
...
+  ///
+  /// In this example, `uniqued()` is used to include only names which are
+  /// unique
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "dog","cat"]
+  ///     let uniqued = animals.uniqued()
+  ///     print(uniqued)
+  ///     // Prints "["dog", "pig","cat", "ox"]"`
+  ///
+  /// - Returns: Returns an array with only the unique elements of this sequence
```
natecook1000, line 28:
```suggestion
  /// - Returns: An array with only the unique elements of this sequence.
```

Sources/Algorithms/Unique.swift
```diff
@@ -14,33 +14,57 @@
...
+  ///  .
+  /// - Complexity: O(*n*), where *n* is the length of the sequence.
   public func uniqued() -> [Element] {
     var seen: Set<Element> = []
     var result: [Element] = []
-    for element in self {
-      if seen.insert(element).inserted {
+    for element in self where seen.insert(element).inserted {
```
natecook1000, line 34: We don't need this kind of lateral change, here and below.

natecook1000 (changes requested):
Thanks for filling out this documentation, @Roshankumar350! A few notes below for you.

Sources/Algorithms/Unique.swift
```diff
@@ -30,8 +39,22 @@ extension Sequence where Element: Hashable {
 
 extension Sequence {
   /// Returns an array with the unique elements of this sequence (as determined
-  /// by the given projection), in the order of the first occurence of each
+  /// by the given projection), in the order of the first occurrence of each
   /// unique element.
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "dog","cat"]
+  ///     let uniqued = animals(on: { $0.first })
```
xwu, line 46: This does not compile.

Sources/Algorithms/Unique.swift
```diff
@@ -30,8 +39,22 @@ extension Sequence where Element: Hashable {
...
+  ///     print(uniqued)
+  ///     // Prints "["dog", "pig","cat", "ox"]"
```
xwu, line 48:
```suggestion
  ///     // Prints '["dog", "pig", "cat", "ox"]'
```

Although the output depends on what the projection chosen for this example is.

Sources/Algorithms/Unique.swift
```diff
@@ -30,8 +39,22 @@ extension Sequence where Element: Hashable {
...
+  ///
+  /// - Parameter projection: A projecting closure. `projection` accepts an
+  ///   element of this sequence as its parameter and returns a projected
+  ///   value of the same  type.
```
xwu, line 52:
```suggestion
  ///   value of the same type.
```

This description is true, but it’s also just describing the type signature in words. But what are the semantics expected of this closure? That should be documented here.

Roshankumar350: Added constraint of "Hashable" element

Sources/Algorithms/Unique.swift
Roshankumar350: Addressed and tested the same example on compiler.

Sources/Algorithms/Unique.swift
```diff
@@ -42,14 +42,14 @@ extension Sequence {
   /// by the given projection), in the order of the first occurrence of each
   /// unique element.
   ///
-  ///     let animals = ["dog", "pig", "cat", "ox", "dog","cat"]
-  ///     let uniqued = animals(on: { $0.first })
+  ///     let animals = ["dog", "pig", "cat", "ox", "dog", "cat"]
```
xwu, line 45: This example doesn't really demonstrate the role of the 'projection' closure. If you want to stick to the projection `$0.first`, I'd suggest something like:

```suggestion
  ///     let animals = ["dog", "pig", "cat", "ox", "cow", "owl"]
```

Sources/Algorithms/Unique.swift
```diff
@@ -42,14 +42,14 @@ extension Sequence {
...
+  ///     let uniqued = animals.uniqued(on: {$0.first})
   ///     print(uniqued)
-  ///     // Prints "["dog", "pig","cat", "ox"]"
+  ///     // Prints '["dog", "pig", "cat", "ox"]'
   ///
   /// - Parameter projection: A projecting closure. `projection` accepts an
   ///   element of this sequence as its parameter and returns a projected
-  ///   value of the same  type.
+  ///   value of the same type having constraint `Hashable`.
```
xwu, line 52: This is not correct: the projected value does *not* need to be of the same type.

But moreover, even when corrected, this statement is still just describing the type of the parameter in words, which is not necessary and doesn't add anything. The challenge here is to teach the user why and how to use a projection function to determine uniqueness.

Sources/Algorithms/Unique.swift
```diff
@@ -30,8 +39,23 @@ extension Sequence where Element: Hashable {
...
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "cow", "owl"]
+  ///     let uniqued = animals.uniqued(on: {$0.first})
+  ///     print(uniqued)
+  ///     // Prints '["dog", "pig", "cat", "ox"]'
+  ///
+  /// - Parameter projection: A projecting closure. `projection` accepts an
+  ///   element of this sequence as its parameter which is having the type of
+  ///   projecting element and returns a projected value that may have the same
+  ///   type having constraint `Hashable`.
```
xwu, line 53: @natecook1000 Do you have suggestions of how to approach describing this parameter?

natecook1000, line 53: How about this?

```suggestion
  /// - Parameter projection: A closure that transforms an element into the
  ///   value to use for uniqueness. If `projection` returns the same value
  ///   for two different elements, the second element will be excluded
  ///   from the resulting array.
```

Sources/Algorithms/Unique.swift
```diff
@@ -30,8 +39,23 @@ extension Sequence where Element: Hashable {
 
 extension Sequence {
   /// Returns an array with the unique elements of this sequence (as determined
-  /// by the given projection), in the order of the first occurence of each
+  /// by the given projection), in the order of the first occurrence of each
   /// unique element.
+  ///
+  ///     let animals = ["dog", "pig", "cat", "ox", "cow", "owl"]
+  ///     let uniqued = animals.uniqued(on: {$0.first})
```
natecook1000, line 46: Let's give an introduction to this example, as well:

```suggestion
  /// This example finds the elements of the `animals` array with unique
  /// first characters:
  ///
  ///     let animals = ["dog", "pig", "cat", "ox", "cow", "owl"]
  ///     let uniqued = animals.uniqued(on: { $0.first })
```

natecook1000 (commented):
Getting close, @Roshankumar350!

Sources/Algorithms/Unique.swift
Roshankumar350: @natecook1000 Thanks for forward thinking.

xwu:
This looks great!

Final step: looks like there's a conflict between this change and the main branch, and the revisions here are best as one commit instead of 18. So please squash and rebase this PR. Here's an introductory guide to doing that from the folks at Servo:

https://github.com/servo/servo/wiki/Beginner%27s-guide-to-rebasing-and-squashing

Sources/Algorithms/Unique.swift
```diff
@@ -15,18 +15,50 @@
...
+    return result
   }
 }
 
 extension Sequence {
   /// Returns an array with the unique elements of this sequence (as determined
-  /// by the given projection), in the order of the first occurence of each
+  /// by the given projection), in the order of the first occurrence of each
   /// unique element.
-  @inlinable
```
xwu, line 29: I don’t think this was rebased correctly. These annotations shouldn’t be deleted.

natecook1000:
Looks great, @Roshankumar350 — thanks for sticking with this! 🎉

PR #40, opened 2020-11-02
Improve performance of Product2.distance(from:to:)

timvermeulen:
This PR gets rid of any unnecessary computations in `Product2.distance(from:to:)`. In particular, it now only does a single pass over `base2` in the worst case (instead of 3), and for non-random-access collections it no longer does more index calculations than the default implementation of `distance(from:to:)`.

### Checklist


natecook1000:
Looks great! And thanks for including the illustrative comments 🖼️

PR #42, opened 2020-11-08
Add random-access methods to SlidingWindows

timvermeulen:
`index(_:offsetBy:)`, `index(_:offsetBy:limitedBy:)`, and `distance(from:to:)` for `SlidingWindows`.

The `base.endIndex..<base.endIndex` representation of `endIndex` made this a bit tricky, because neither of the bounds is a logical successor of the index before it. The `offsetForward`/`offsetBackward` dance alleviates this somewhat by ensuring that the limit is always on the correct side of `i` and never equal to it, removing some of the edge cases.

### Checklist

### DIFF
--- a/Sources/Algorithms/SlidingWindows.swift
+++ b/Sources/Algorithms/SlidingWindows.swift
@@ -84,9 +84,198 @@ extension SlidingWindows: Collection {
     )
   }
   
-  // TODO: Implement distance(from:to:), index(_:offsetBy:) and
-  // index(_:offsetBy:limitedBy:)
+  public func index(_ i: Index, offsetBy distance: Int) -> Index {
+    guard distance != 0 else { return i }
+    
+    return distance > 0
+      ? offsetForward(i, by: distance)
+      : offsetBackward(i, by: -distance)
+  }
+  
+  public func index(
+    _ i: Index,
+    offsetBy distance: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    guard distance != 0 else { return i }
+    guard limit != i else { return nil }
+    
+    if distance > 0 {
+      return limit > i
+        ? offsetForward(i, by: distance, limitedBy: limit)
+        : offsetForward(i, by: distance)
+    } else {
+      return limit < i
+        ? offsetBackward(i, by: -distance, limitedBy: limit)
+        : offsetBackward(i, by: -distance)
+    }
+  }
+  
+  private func offsetForward(_ i: Index, by distance: Int) -> Index {
+    offsetForward(i, by: distance, limitedBy: endIndex)!
+  }
+  
+  private func offsetBackward(_ i: Index, by distance: Int) -> Index {
+    offsetBackward(i, by: distance, limitedBy: startIndex)!
+  }
+  
+  private func offsetForward(
+    _ i: Index, by distance: Int, limitedBy limit: Index
+  ) -> Index? {
+    assert(distance > 0)
+    assert(limit > i)
+    
+    // `endIndex` and the index before it both have `base.endIndex` as their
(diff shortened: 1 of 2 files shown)


ollieatkinson:
> The base.endIndex..<base.endIndex representation of endIndex made this a bit tricky

Do you think it's worth considering a better implementation of `endIndex` in this case?

timvermeulen:
>Do you think it's worth considering a better implementation of `endIndex` in this case?

That's always worth considering, but I don't think there is any! If we want `endIndex` to fit in naturally with the other indices, its lower and upper bounds would need to be the direct successors of the bounds of the index before `endIndex`. But since the index before it corresponds to the last window which has an upper bound of `base.endIndex`, there is no successor to that (in the general case).

Of course we could give `SlidingWindows.Index` an internal enum representation with an `endIndex` case, which I'm mostly indifferent towards, but that wouldn't get rid of any edge cases. It might make it less error-prone to work with, but with the tests passing I'm pretty confident this implementation is correct as well.

Having said all that, it's totally possible to write a `protocol ValidIndexCollection: Collection { ... }` that lets you write things like

```swift
struct CollectionOfTwo<Element> {
  let first, second: Element
}

extension CollectionOfTwo: ValidIndexCollection {
  enum ValidIndex: Comparable {
    case first, second
  }
  
  func element(at index: ValidIndex) -> Element {
    index == .first ? first : second
  }
  
  var firstValidIndex: ValidIndex? { .first }
  var lastValidIndex: ValidIndex? { .second }
  
  func validIndex(after index: ValidIndex) -> ValidIndex {
    .second
  }
}
```

where `ValidIndexCollection` handles all the `endIndex` edge cases for you. This could be useful for implementing collections that don't have a natural way of representing `endIndex`. `SlidingWindows` and `ClosedRange` come to mind. This could be worth considering if we end up with other collections that need to deal with `endIndex` separately.

Sources/Algorithms/SlidingWindows.swift
```diff
@@ -84,9 +84,198 @@ extension SlidingWindows: Collection {
...
+        : offsetForward(i, by: distance)
+    } else {
+      return limit < i
+        ? offsetBackward(i, by: -distance, limitedBy: limit)
+        : offsetBackward(i, by: -distance)
+    }
+  }
+  
+  private func offsetForward(_ i: Index, by distance: Int) -> Index {
+    offsetForward(i, by: distance, limitedBy: endIndex)!
```
ollieatkinson, line 115: Instead of force unwrapping here (and `offsetBackward`) I would add an error message describing why the offset was not valid. We did something similar [here](https://github.com/apple/swift-algorithms/pull/42/files#diff-2d7e1f0b61f31e10e56c985d329b222f086037d37960895c16669eaedd036515R79)

Sources/Algorithms/SlidingWindows.swift
```diff
@@ -84,9 +84,198 @@ extension SlidingWindows: Collection {
...
+    // `endIndex` and the index before it both have `base.endIndex` as their
+    // upper bound, so we first advance to the base index _before_ the upper
+    // bound of the output, in order to avoid advancing past the end of `base`
+    // when advancing to `endIndex`.
+    //
+    // Advancing by 4:
+    //
+    //  input: [x|x x x x x|x x x x]        [x x|x x x x x|x x x]
+    //                     |> > >|>|   or                 |> > >|
+    // output: [x x x x x|x x x x x]        [x x x x x x x x x x]  (`endIndex`)
```
ollieatkinson, line 137: 👏 This is excellent and a massive help when reviewing this code

natecook1000:
Thanks, @timvermeulen! 👏

PR #43, opened 2020-11-13
Merging Sorted Sequences

CTMacUser:
This library adapts the C++ algorithm functions that take two sorted sequences and returns an also-sorted sequence that represents the set-difference, intersection, union, *etc.*. Instead of separate functions for each operation, the desired operation is selected by an enumeration value.  (All the operations follow the same algorithm.) This includes an operation to do the general merge sort!

### Checklist

### DIFF
--- a/Guides/MergeSorted.md
+++ b/Guides/MergeSorted.md
@@ -1,0 +1,125 @@
+# Permutations
+
+[[Source](../Sources/Algorithms/MergeSorted.swift) | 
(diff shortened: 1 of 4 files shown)


Guides/MergeSorted.md
```diff
@@ -0,0 +1,125 @@
+# Permutations
```
mdznr, line 1:
```suggestion
# MergeSorted
```

CTMacUser:
This is now obsolete from my next attempt at [#184].

PR #47, opened 2020-11-25
Add conformances to LazySequenceProtocol and LazyCollectionProtocol

danielctull:
This fixes an omission in `Intersperse` where it doesn't currently conform to `LazySequenceProtocol` when the base sequence also conforms to `LazySequenceProtocol`.

### Checklist

### DIFF
--- a/Guides/Indexed.md
+++ b/Guides/Indexed.md
@@ -30,6 +30,6 @@ extension Collection {
 ```
 
 `Indexed` scales from a collection up to a random-access collection, depending on 
-its base type. `Indexed` also conforms to `LazySequenceProtocol` when the base type
-conforms.
+its base type. `Indexed` also conforms to `LazySequenceProtocol` and 
+`LazyCollectionProtocol` when the base type conforms.
 
--- a/Guides/Intersperse.md
+++ b/Guides/Intersperse.md
@@ -30,9 +30,9 @@ extension Sequence {
 ```
 
 The new `Intersperse` type represents the sequence when the separator is
-inserted between each element. Intersperse conforms to Collection and 
-BidirectionalCollection when the base sequence conforms to Collection and
-BidirectionalCollection respectively.
+inserted between each element. `Intersperse` conforms to `Collection`,
+`BidirectionalCollection`, `RandomAccessCollection`, and `LazySequenceProtocol`
+when the base sequence conforms to those respective protocols.
 
 ### Complexity
 
--- a/Guides/Intersperse.md
+++ b/Guides/Intersperse.md
@@ -31,8 +31,9 @@ extension Sequence {
 
 The new `Intersperse` type represents the sequence when the separator is
 inserted between each element. `Intersperse` conforms to `Collection`,
-`BidirectionalCollection`, `RandomAccessCollection`, and `LazySequenceProtocol`
-when the base sequence conforms to those respective protocols.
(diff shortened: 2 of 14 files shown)


CTMacUser:
Probably should add (conditional) `LazyCollectionProtocol` support too.  (It's actually a separate refined protocol, not a `typealias` of `LazySequenceProtocol` for collections.)

timvermeulen:
Huh, I was under the impression that `LazyCollectionProtocol` was a type alias when I added `LazySequenceProtocol` conformances in #29. Other collections in this library could benefit from a `LazyCollectionProtocol` conformance too, then.

The `LazyCollectionProtocol` docs say
>To add new lazy collection operations, extend this protocol with methods that return lazy wrappers that are themselves `LazyCollectionProtocol`s.
so we indeed can't assume that `LazySequenceProtocol + Collection` conformance is sufficient.

CTMacUser:
I was surprised too when I re-read the `LazyCollectionProtocol` definition.

danielctull:
I've just been through all the types and the only one that can be conformed to `LazyCollectionProtocol` was `Indexed` as the rest aren't collections, they can't conform to it. 

I have renamed the `XCTAssertLazy` function to `XCTAssertLazySequence` and added a similar one called `XCTAssertLazyCollection` and fixed up the `Indexed` type too.

danielctull:
> I've just been through all the types and the only one that can be conformed to `LazyCollectionProtocol` was `Indexed` as the rest aren't collections, they can't conform to it.

I realised that I was only looking at types that already conform to `LazySequenceProtocol`, the following three are other types in the package that feel like they could conditionally conform to `LazySequenceProtocol` and/or `LazyCollectionProtocol `.

* `Chain`
* `Product2`
* `SlidingWindows`

timvermeulen:
`Chain2` and `Product2` don't need the `LazySequenceProtocol` conformance because they're created by the `chain` and `product` free functions, not by any methods that can be part of a (lazy) chain.

`SlidingWindows` does indeed still need the conformance though!

danielctull:
> `Chain2` and `Product2` don't need the `LazySequenceProtocol` conformance because they're created by the `chain` and `product` free functions, not by any methods that can be part of a (lazy) chain.

It would still be beneficial for them to have the conformance though? The following works if `Chain2` conforms to `LazySequenceProtocol`, otherwise you have to go through another `lazy` operator.

``` swift
extension LazySequenceProtocol {
  func someLazyOnlyFunction() {}
}

let first = (1...).prefix(10).lazy
let second = (20...).lazy
chain(first, second).someLazyOnlyFunction()
```

timvermeulen:
>It would still be beneficial for them to have the conformance though?

In theory nothing is stopping them from conforming, but the laziness of sequences is only expected to propagate through method calls (lazy chaining) and not through the parameters of functions like `zip`/`chain`/`product`. It's just convention 🤷

It could also be ambiguous whether both input sequences would need to be lazy for the laziness to propagate, or that just one is enough. Always propagating the laziness of the `Self` type clears up this ambiguity. (Note how `Sequence.flatMap` propagates the laziness of the original sequence, and not the laziness of the sequence returned from the closure.)

danielctull:
Thanks for explaining that @timvermeulen.

In that case I believe this PR is ready to go. 🙂

natecook1000:
Looks great — thanks so much for noticing and fixing this, @danielctull!

Issue #23, opened 2020-10-14
Add `indexBeforeFirst(where:)` and `indexAfterLast(where:)`


kylemacomber:
In #4 the [implementation](https://github.com/apple/swift-algorithms/pull/4/files#diff-98b27dd842538ebcf96bc8d197b4de38b8eddb255720be513ae70b1ab4356e86R37) of `trimmed(where:)` requires a raw loop because using `lastIndex(where:)` would require making an extra call to `index(after:)` to advanced the returned index.

The [embracing algorithms](https://developer.apple.com/videos/play/wwdc2018/223/) talk exhibits the need for the dual—the index before `firstIndex(where:)`—to avoid a raw loop or an extra call to `index(before:)`:

```swift
extension MutableCollection {
    mutating func bringForward(elementsSatisfying predicate: (Element) -> Bool) {
        if let predecessor = indexBeforeFirst(where: predicate) {
            self[predecessor...].stablePartition(isSuffixElement: { !predicate($0) })
        }
    }
}

extension Collection {
    func indexBeforeFirst(where predicate: (Element) -> Bool) -> Index? {
        return indices.first {
            let successor = index(after: $0)
            return successor != endIndex && predicate(self[successor])
        }
    } 
}
```

These use cases seem like sufficient justification to add these algorithms.

Though I wonder if `indexBeforeFirst(where:)`/`indexAfterLast(where:)` or `firstIndex(before:)`/`lastIndex(after:)` or some other naming scheme is best?

natecook1000:
The naming for these is tricky! It would be nice if they sorted next to `firstIndex(where:)` and `lastIndex(where:)`, but the contortions to make that happen don't seem worth it. When used with a trailing closure, `firstIndex(before:)` becomes `firstIndex { ... }`, which loses a vital bit of meaning. `firstIndexBefore(where:)` is ungainly, and even `firstIndexBefore(_:)` doesn't seem quite right.

Let's stick with the clear `indexBeforeFirst(where:)` and `indexAfterLast(where:)`.

xwu:
In fact, the “firstness” and “lastness” of these APIs are kind of inconsequential, and even potentially misinterpretable. The index _before the first x_ is actually the _last index before x_.

[...]

PR #50, opened 2020-12-18
Rename `trimming(where:)` to `trimming(while:)`

benrimmington:
The `where:` argument label is more commonly used:

* `Sequence.contains(where:)`
* `Sequence.first(where:)`
* `Collection.firstIndex(where:)`
* `BidirectionalCollection.last(where:)`
* `BidirectionalCollection.lastIndex(where:)`
* `RangeReplaceableCollection.removeAll(where:)`

But the `while:` argument label is used for "consecutive elements that satisfy the given predicate":

* `Sequence.drop(while:)`
* `Sequence.prefix(while:)`

### Checklist

### DIFF
--- a/Guides/Trim.md
+++ b/Guides/Trim.md
@@ -6,13 +6,13 @@
 Returns a `SubSequence` formed by discarding all elements at the start and end of the collection
 which satisfy the given predicate.
 
-This example uses `trimming(where:)` to get a substring without the white space at the beginning and end of the string.
+This example uses `trimming(while:)` to get a substring without the white space at the beginning and end of the string.
 
 ```swift
 let myString = "   hello, world  "
-print(myString.trimming(where: \.isWhitespace)) // "hello, world"
+print(myString.trimming(while: \.isWhitespace)) // "hello, world"
 
-let results = [2, 10, 11, 15, 20, 21, 100].trimming(where: { $0.isMultiple(of: 2) })
+let results = [2, 10, 11, 15, 20, 21, 100].trimming(while: { $0.isMultiple(of: 2) })
 print(results) // [11, 15, 20, 21]
 ```
 
@@ -23,7 +23,7 @@ A new method is added to `BidirectionalCollection`:
 ```swift
 extension BidirectionalCollection {
 
-  public func trimming(where predicate: (Element) throws -> Bool) rethrows -> SubSequence
+  public func trimming(while predicate: (Element) throws -> Bool) rethrows -> SubSequence
 }
 ```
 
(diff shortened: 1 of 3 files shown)


benrimmington:
Cc: @karwa, @natecook1000

kylemacomber:
I think this makes sense. Good observation @benrimmington!

timvermeulen:
I think the idea behind the current naming is that the word "trimming" already implies that only consecutive elements at the start or end of the collection are excluded. The `while:` label is probably not strictly necessary to convey this information, although I might be biased by having used similarly named functionality in other languages.

A possible downside of the `while:` label is that it might give someone the impression that only a prefix is excluded. If you go through the collection in the forward direction and only trim elements "while" the predicate is true, you never reach the end. The existing `drop(while:)` and `prefix(while:)` both operate on a prefix of the receiver.

benrimmington:
@timvermeulen The `while:` argument label makes more sense if we also have:

* `trimmingPrefix(while:)` — instead of `drop(while:)`
* `trimmingSuffix(while:)`

The implementation is easier with `indexAfterLast(where:)` from issue #23.

```swift

  public func trimming(
  ) rethrows -> SubSequence {
    guard
      let fromIndex = try firstIndex(where: { try predicate($0) == false }),
      let upToIndex = try indexAfterLast(where: { try predicate($0) == false }),
      fromIndex <= upToIndex
    else {
      return suffix(0)
    }
    return self[fromIndex..<upToIndex]
  }

  public func trimmingPrefix(
  ) rethrows -> SubSequence {
    guard
      let fromIndex = try firstIndex(where: { try predicate($0) == false })
    else {
      return suffix(0)
    }
    return self[fromIndex...]
  }

  public func trimmingSuffix(
  ) rethrows -> SubSequence {
    guard
      let upToIndex = try indexAfterLast(where: { try predicate($0) == false })
    else {
      return prefix(0)
    }
    return self[..<upToIndex]
  }
}
```

natecook1000:
Thanks for opening this PR/discussion, @benrimmington! 

I think @timvermeulen identifies a valid conflict. If we mean for `while` to imply that a predicate will match a contiguous range of elements, then trimming from both sides doesn't quite qualify, since it can match _two_ ranges, even if `trimmingPrefix(while:)` and `trimmingSuffix(while:)` do. (I'd like to add those additional methods eventually, as well as the missing `suffix(while:)`.)

So the question boils down to whether we should value consistency across method names over this kind of technical accuracy, and if so, which side we should lean toward. In this case, I think there's value in not making users jump through the hoops of understanding this fairly subtle distinction, and once all three trimming methods are present, I don't think there will be much cause for confusion about what the bare `trimming` method does. Given that `prefix(while:)` isn't going anywhere, and seems well named, let's go with `while` for the whole lot.

karwa:
The word `while` has a kind of order or temporal connotation to me. I think of “trimming while (some condition)” as meaning that it only trims the prefix, and stops as soon as the “while” condition is not true.

`where` doesn’t have that same connotation in my mind, or at least it’s not as strong as it is for `while`.

Just MHO.

natecook1000:
Merged manually — thanks, @benrimmington! 🎉

PR #51, opened 2020-12-18
All combinations

mdznr:
### Description
This addition adds the ability to easily get combinations of a range of sizes, instead of the previous behavior of a single, fixed size, `k`.

### Detailed Design
The `combinations(ofCount:)` function that takes a `RangeExpression` is added in an extension of `Collection`.

```swift
extension Collection {
  public func combinations<R: RangeExpression>(
    ofCount kRange: R
  ) -> Combinations<Self> where R.Bound == Int {
    return Combinations(self, kRange: kRange)
  }
}
```

If the upper bound of the range exceeds the length of the collection, no such combinations of those sizes will be returned, effectively clamping the range to `0...n` where `n` is the length of the collection. The behavior of negative numbers also remains unchanged (asserting).

This also updates the implementation of the existing `combinations(ofCount:)` function to initialize a `Combinations` with a range of `k...k` to provide the same functionality as it did before.

```swift
public struct Combinations<Base: Collection> {
  internal init(_ base: Base, k: Int) {
    self.init(base, kRange: k...k)
  }
 }
```

### Complexity
The complexity remains O(*1*) for initializing `Combinations`. Accessing the next combination also remains O(*1*).

### Naming
The name `combinations(ofCount:)` is the same as the original, but reflects the ability to provide multiple counts. Alternatively considered `combinations(ofCounts:)` and `combinations(ofCountsIn range:)`.

### Documentation Plan
The documentation in `Guides/Combinations.md` has been updated to reflect the new function. Inline code documentation also exists for each of the functions.

### Test Plan
- Adds tests for the `count` property, which was previously untested
- Adds tests for the outputs of a range of sizes
- Adds tests for all valid sizes
- Adds tests for ranges extending above the size of the collection

### Source Impact
The functionality is purely additive. The behavior of the existing function is unchanged.

### Checklist

### DIFF
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -239,8 +239,8 @@ extension Collection {
   ///     // [10, 20, 30, 40]
   ///
   /// If `kRange` is `0...0`, the resulting sequence has exactly one element, an
-  /// empty array. If `k.upperBound` is greater than the number of elements in
-  /// this sequence, the resulting sequence has no elements.
+  /// empty array. The given range is limited to `0...base.count`. If the
+  /// limited range is empty, the resulting sequence has no elements.
   ///
   /// - Parameter kRange: The range of numbers of elements to include in each
   /// combination.
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -15,7 +15,7 @@ public struct Combinations<Base: Collection> {
   public let base: Base
   
   @usableFromInline
-  internal var k: Int
+  internal let k: Int
   
   @usableFromInline
   internal init(_ base: Base, k: Int) {
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -14,6 +14,9 @@ public struct Combinations<Base: Collection> {
   /// The collection to iterate over for combinations.
   public let base: Base
   
+  @usableFromInline
+  internal let baseCount: Int
+  
   /// The range of accepted sizes of combinations.
   /// - Note: This may be `nil` if the attempted range entirely exceeds the
   /// upper bounds of the size of the `base` collection.
@@ -40,9 +43,11 @@ public struct Combinations<Base: Collection> {
   ) where R.Bound == Int {
     let range = kRange.relative(to: 0 ..< .max)
     self.base = base
-    let upperBound = base.count + 1
+    let baseCount = base.count
+    self.baseCount = baseCount
+    let upperBound = baseCount + 1
     self.kRange = range.lowerBound < upperBound
-      ? range.clamped(to: 0..<upperBound)
+      ? range.clamped(to: 0 ..< upperBound)
       : nil
   }
   
@@ -50,8 +55,8 @@ public struct Combinations<Base: Collection> {
   @inlinable
   public var count: Int {
     guard let k = self.kRange else { return 0 }
-    let n = base.count
-    if k == 0..<(n + 1) {
+    let n = baseCount
+    if k == 0 ..< (n + 1) {
       return 1 << n
     }
     
@@ -123,7 +128,7 @@ extension Combinations: Sequence {
       func advanceKRange() {
         if kRange.lowerBound < kRange.upperBound {
           let advancedLowerBound = kRange.lowerBound + 1
-          kRange = advancedLowerBound..<kRange.upperBound
+          kRange = advancedLowerBound ..< kRange.upperBound
           indexes.removeAll(keepingCapacity: true)
           indexes.append(contentsOf: base.indices.prefix(kRange.lowerBound))
         }
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -122,7 +122,7 @@ extension Combinations: Sequence {
     internal mutating func advance() {
       /// Advances `k` by increasing its `lowerBound` or finishes the iteration.
       func advanceK() {
-        let advancedLowerBound = k.lowerBound.advanced(by: 1)
+        let advancedLowerBound = k.lowerBound + 1
         if advancedLowerBound < k.upperBound {
           k = advancedLowerBound..<k.upperBound
           self.indexes = Array(base.indices.prefix(k.lowerBound))
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -177,7 +177,7 @@ extension Combinations: Equatable where Base: Equatable {}
 extension Combinations: Hashable where Base: Hashable {}
 
 //===----------------------------------------------------------------------===//
-// combinations(ofCounts:)
+// combinations(ofCount:)
 //===----------------------------------------------------------------------===//
 
 extension Collection {
@@ -188,7 +188,7 @@ extension Collection {
   /// four colors:
   ///
   ///     let colors = ["fuchsia", "cyan", "mauve", "magenta"]
-  ///     for combo in colors.combinations(ofCounts: 1...2) {
+  ///     for combo in colors.combinations(ofCount: 1...2) {
   ///         print(combo.joined(separator: ", "))
   ///     }
   ///     // fuchsia
@@ -242,17 +242,11 @@ extension Collection {
   /// - Complexity: O(1)
   @inlinable
   public func combinations<R: RangeExpression>(
-    ofCounts kRange: R
+    ofCount kRange: R
   ) -> Combinations<Self> where R.Bound == Int {
     return Combinations(self, kRange: kRange)
   }
-}
-
-//===----------------------------------------------------------------------===//
-// combinations(ofCount:)
-//===----------------------------------------------------------------------===//
-
-extension Collection {
+  
   /// Returns a collection of combinations of this collection's elements, with
   /// each combination having the specified number of elements.
   ///
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -14,18 +14,49 @@ public struct Combinations<Base: Collection> {
   /// The collection to iterate over for combinations.
   public let base: Base
   
+  /// The range of accepted sizes of combinations.
+  /// - Note: This may be `nil` if the attempted range entirely exceeds the
+  /// upper bounds of the size of the `base` collection.
   @usableFromInline
-  internal let k: Int
+  internal let k: ClosedRange<Int>?
   
+  /// Initializes a `Combinations` for all combinations of `base` of all sizes.
+  /// - Parameter base: The collection to iterate over for combinations.
+  @usableFromInline
+  internal init(_ base: Base) {
+    self.init(base, k: 0...base.count)
+  }
+  
+  /// Initializes a `Combinations` for all combinations of `base` of size `k`.
+  /// - Parameters:
+  ///   - base: The collection to iterate over for combinations.
+  ///   - k: The expected size of each combination.
   @usableFromInline
   internal init(_ base: Base, k: Int) {
+    self.init(base, k: k...k)
+  }
+  
+  /// Initializes a `Combinations` for all combinations of `base` of sizes
+  /// within a given range.
+  /// - Parameters:
+  ///   - base: The collection to iterate over for combinations.
+  ///   - k: The range of accepted sizes of combinations.
+  @usableFromInline
+  internal init(_ base: Base, k: ClosedRange<Int>) {
+    assert(k.lowerBound >= 0, "Can't have combinations with a negative number of elements.")
     self.base = base
-    self.k = base.count < k ? -1 : k
+    self.k = (k.lowerBound <= base.count) ? k.clamped(to: 0...base.count) : nil
   }
   
   /// The total number of combinations.
   @inlinable
   public var count: Int {
+    guard let k = self.k else { return 0 }
+    let n = base.count
+    if k == 0...n {
+      return 1 << n
+    }
+    
     func binomial(n: Int, k: Int) -> Int {
       switch k {
       case n, 0: return 1
@@ -35,9 +66,9 @@ public struct Combinations<Base: Collection> {
       }
     }
     
-    return k >= 0
-      ? binomial(n: base.count, k: k)
-      : 0
+    return k.map {
+      binomial(n: n, k: $0)
+    }.reduce(0, +)
   }
 }
 
@@ -47,6 +78,10 @@ extension Combinations: Sequence {
     @usableFromInline
     internal let base: Base
     
+    /// The current range of accepted sizes of combinations.
+    @usableFromInline
+    internal var k: ClosedRange<Int>
+    
     @usableFromInline
     internal var indexes: [Base.Index]
     
@@ -55,10 +90,9 @@ extension Combinations: Sequence {
     
     internal init(_ combinations: Combinations) {
       self.base = combinations.base
-      self.finished = combinations.k < 0
-      self.indexes = combinations.k < 0
-        ? []
-        : Array(combinations.base.indices.prefix(combinations.k))
+      self.k = combinations.k ?? 0...0
+      self.indexes = Array(combinations.base.indices.prefix(k.upperBound))
+      self.finished = (combinations.k == nil)
     }
     
     /// Advances the current indices to the next set of combinations. If
@@ -88,17 +122,22 @@ extension Combinations: Sequence {
         finished = true
         return
       }
-    
+      
       let i = indexes.count - 1
       base.formIndex(after: &indexes[i])
       if indexes[i] != base.endIndex { return }
-
+      
       var j = i
       while indexes[i] == base.endIndex {
         j -= 1
         guard j >= 0 else {
-          // Finished iterating over combinations
-          finished = true
+          // Finished iterating over combinations of this size.
+          if k.lowerBound < k.upperBound {
+            k = k.lowerBound...k.upperBound.advanced(by: -1)
+            self.indexes = Array(base.indices.prefix(k.upperBound))
+          } else {
+            finished = true
+          }
           return
         }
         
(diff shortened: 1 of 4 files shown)


natecook1000 (changes requested):
Thanks for this addition, @mdznr! 👏👏

I'm a big fan of the range-based version, especially that you can pass all kinds of range expressions and it just works. The double plural in the name strikes me as a _little_ odd, but I don't think it's worth changing unless there's something clearly better.

I'm not sold on the no-argument version. Can you explain a bit about your use case for including this? As a user, I would be surprised that the result contained single-element "combinations" and _very_ surprised about the leading zero-element value. I'm inclined to just document the usage in the range-based version, since you can use a range expression to explicitly get the behavior you want — e.g. `numbers.combinations(ofCounts: 0...)`.

Sources/Algorithms/Combinations.swift
```diff
@@ -14,29 +14,66 @@ public struct Combinations<Base: Collection> {
...
     func binomial(n: Int, k: Int) -> Int {
       switch k {
-      case n, 0: return 1
-      case n...: return 0
-      case (n / 2 + 1)...: return binomial(n: n, k: n - k)
-      default: return n * binomial(n: n - 1, k: k - 1) / k
+        case n, 0: return 1
+        case n...: return 0
+        case (n / 2 + 1)...: return binomial(n: n, k: n - k)
+        default: return n * binomial(n: n - 1, k: k - 1) / k
```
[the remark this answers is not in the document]
mdznr: Thanks for clarifying! I had just used **Editor > Structure > Re-Indent** on the file, but looks like the preferences for files in the sidebar don’t have the ability to adjust this behavior.

Sources/Algorithms/Combinations.swift
```diff
@@ -129,7 +180,100 @@ extension Combinations: Equatable where Base: Equatable {}
...
+  ///     // fuchsia, magenta
+  ///     // cyan, mauve
+  ///     // cyan, magenta
+  ///     // mauve, magenta
+  ///
+  /// The returned collection presents combinations in a consistent order, where
+  /// the indices in each combination are in ascending lexicographical order.
+  /// That is, in the example above, the combinations in order are the elements
+  /// at `[0, 1]`, `[0, 2]`, `[0, 3]`, `[1, 2]`, `[1, 3]`, `[2, 3]`, `[0]`,
+  /// `[1]`, `[2]`, and finally `[3]`.
```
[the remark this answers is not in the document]
mdznr: It did! Thanks for catching this.

Sources/Algorithms/Combinations.swift
```diff
@@ -14,29 +14,66 @@ public struct Combinations<Base: Collection> {
...
+  /// Initializes a `Combinations` for all combinations of `base` of sizes
+  /// within a given range.
+  /// - Parameters:
+  ///   - base: The collection to iterate over for combinations.
+  ///   - k: The range of accepted sizes of combinations.
+  @usableFromInline
+  internal init<R: RangeExpression>(
+    _ base: Base, k: R
+  ) where R.Bound == Int {
+    let range = k.relative(to: R.Bound.zero..<R.Bound.max)
```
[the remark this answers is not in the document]
mdznr: 3b2a25d

Sources/Algorithms/Combinations.swift
```diff
@@ -14,29 +14,66 @@ public struct Combinations<Base: Collection> {
...
+        default: return n * binomial(n: n - 1, k: k - 1) / k
       }
     }
     
-    return k >= 0
-      ? binomial(n: base.count, k: k)
-      : 0
+    return k.map {
+      binomial(n: n, k: $0)
+    }.reduce(0, +)
```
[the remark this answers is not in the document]
mdznr: Good call. I’ll play around with a way to do this that maintains readability. We can probably get most of the algorithmic performance benefits by using memoization in `binomial`.

Sources/Algorithms/Combinations.swift
```diff
@@ -14,29 +14,66 @@ public struct Combinations<Base: Collection> {
   /// The collection to iterate over for combinations.
   public let base: Base
   
+  /// The range of accepted sizes of combinations.
+  /// - Note: This may be `nil` if the attempted range entirely exceeds the
+  /// upper bounds of the size of the `base` collection.
   @usableFromInline
-  internal var k: Int
+  internal let k: Range<Int>?
```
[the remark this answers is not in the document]
mdznr: Perhaps `kRange`? Just don’t want it to potentially get confused with the `count` property.

Sources/Algorithms/Combinations.swift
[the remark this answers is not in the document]
mdznr: I tested adding a small memoization cache (`Dictionary`) to `binomial`. To get the `count` value for `Combinations` of a `Collection` of length 26 (`"ABCDEFGHIJKLMNOPQRSTUVWXYZ"`), using count `1..<26` it cut down the number of calls to `binomial` from **206** to **128**.

I have some ideas on how to further improve performance for this. There is already a special case when the range of `k` is `0...base.count` to use `1 << n` (2^n). However, a potential common case is excluding the empty and/or complete set (`1..<base.count`), which we can easily calculate using 2^n minus 1 (or 2). We can do some work to get the absolute minimum number of computations necessary to calculate a contiguous range of values in the last row of the [arithmetic triangle](https://en.wikipedia.org/wiki/Pascal%27s_triangle).

Should I move that to a separate, subsequent PR so not to block this one?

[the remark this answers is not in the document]
natecook1000: That can be done later for sure!

mdznr:
Thanks for the feedback, @natecook1000 !

> The double plural in the name strikes me as a _little_ odd, but I don't think it's worth changing unless there's something clearly better.

I agree that it does sound a bit odd. I’d definitely be up for changing it if anyone has any suggestions. Perhaps something like `combinations(ofCountIn kRange: R)`?

> I'm not sold on the no-argument version. Can you explain a bit about your use case for including this? As a user, I would be surprised that the result contained single-element "combinations" and _very_ surprised about the leading zero-element value. I'm inclined to just document the usage in the range-based version, since you can use a range expression to explicitly get the behavior you want — e.g. `numbers.combinations(ofCounts: 0...)`.

I do see why it seems bit weird at first, but there’s a couple reasons why I have it as such:

1. It’s consistent with well-known mathematical topics.<sup>[[1]](https://en.wikipedia.org/wiki/Power_set#Example)</sup><sup>[[2]](https://en.wikipedia.org/wiki/Combination#Number_of_k-combinations_for_all_k)</sup> 
2. It works well with two related additions that I may put up in future subsequent PRs:
  i. **Bifurcate.** This addition would be a function like the standard library’s `filter(_:)` but returns a tuple of two arrays with the _included_ elements in the first, and the _excluded_ in the second. `bifurcate(_:) -> ([E], [E])`
  ii. **BifurcatedCombinations.** This builds off of `Combinations` and `Bifurcate` to allow you to easily iterate over all combinations of a collection. If you were building something like a [matching-type](https://en.wikipedia.org/wiki/Card_game#Matching_games) card game where you draw cards and have to create [melds](https://en.wikipedia.org/wiki/Meld_(cards)), it is useful to go through all possible combinations, including the empty and complete sets. For example, if you have `[7♦, 8♦, 9♦]` and you drew `[5♦, 5♣]` (able to discard/keep any number of them), it would be good to know that you have four options:
    - Keep `[]`; discard `[5♦, 5♣]`; (to maintain a hand complete with melds)
    - Keep `[5♦]`; discard `[5♣]`; (in hopes of drawing `6♦` to create a run of `5-9♦` later)
    - Keep `[5♣]`; discard `[5♦]`; (probably not a great strategy in this game)
    - Keep `[5♦, 5♣]`; discard `[]`; (in hopes of creating a set of 3 or more `5`s later)

kylemacomber:
> > The double plural in the name strikes me as a _little_ odd, but I don't think it's worth changing unless there's something clearly better.
> 
> I agree that it does sound a bit odd. I’d definitely be up for changing it if anyone has any suggestions. Perhaps something like `combinations(ofCountIn kRange: R)`?

We might be able to get away with just overloading `combinations(ofCount:)`:

- `combinations(ofCount: 3..<6)` — "combinations of count 3 up to 6"
- `combinations(ofCount: 3...5)` — "combinations of count 3 through 5"
- `combinations(ofCount: 3...)` — "combinations of count greater than or equal to 3"
- etc.

>> I'm inclined to just document the usage in the range-based version, since you can use a range expression to explicitly get the behavior you want — e.g. `numbers.combinations(ofCounts: 0...)`.
>
> I do see why it seems bit weird at first, but there’s a couple reasons why I have it as such:
> 
> 1. It’s consistent with well-known mathematical topics.[[1]](https://en.wikipedia.org/wiki/Power_set#Example)[[2]](https://en.wikipedia.org/wiki/Combination#Number_of_k-combinations_for_all_k)
> 2. It works well with two related additions that I may put up in future subsequent PRs:

I think these are some strong points, however I'm not sure they overcomes (for me at least) just how trivial and explicit it is to compose `combinations(ofCounts: 0...)` on your own. It seems to me that it might be better to wait and see if `combinations(ofCounts: 0...)` becomes relatively common and idiomatic, and then discuss adding a convenience.

mdznr:
> > > The double plural in the name strikes me as a _little_ odd, but I don't think it's worth changing unless there's something clearly better.
> > 
> > 
> > I agree that it does sound a bit odd. I’d definitely be up for changing it if anyone has any suggestions. Perhaps something like `combinations(ofCountIn kRange: R)`?
> 
> We might be able to get away with just overloading `combinations(ofCount:)`:

I like that idea. I’ll go ahead and make this change.

> 
> > > I'm inclined to just document the usage in the range-based version, since you can use a range expression to explicitly get the behavior you want — e.g. `numbers.combinations(ofCounts: 0...)`.
> > 
> > 
> > I do see why it seems bit weird at first, but there’s a couple reasons why I have it as such:
> > 
> > 1. It’s consistent with well-known mathematical topics.[[1]](https://en.wikipedia.org/wiki/Power_set#Example)[[2]](https://en.wikipedia.org/wiki/Combination#Number_of_k-combinations_for_all_k)
> > 2. It works well with two related additions that I may put up in future subsequent PRs:
> 
> I think these are some strong points, however I'm not sure they overcomes (for me at least) just how trivial and explicit it is to compose `combinations(ofCounts: 0...)` on your own. It seems to me that it might be better to wait and see if `combinations(ofCounts: 0...)` becomes relatively common and idiomatic, and then discuss adding a convenience.

Sounds good 👍. My original use case was to iterate through all combinations, so that’s where I started. Support for arbitrary ranges was added along the way since it was easy to do.

natecook1000:
> > We might be able to get away with just overloading `combinations(ofCount:)`:
>
> I like that idea. However, while having two functions named `combinations(ofCount:)` compiles, lines of code that try to call it doesn’t compile, unfortunately.

That's a great idea, Kyle. @mdznr can you double check your error here? I'm not seeing an ambiguity with that change — maybe you had a stale `Algorithms` module when compiling.

natecook1000:
Also, I've definitely seen requests for `bifurcate` before, but haven't heard of `bifurcatedCombinations`. Interesting!

mdznr:
> > > We might be able to get away with just overloading `combinations(ofCount:)`:
> > 
> > 
> > I like that idea. However, while having two functions named `combinations(ofCount:)` compiles, lines of code that try to call it doesn’t compile, unfortunately.
> 
> That's a great idea, Kyle. @mdznr can you double check your error here? I'm not seeing an ambiguity with that change — maybe you had a stale `Algorithms` module when compiling.

I was surprised to see that it wasn’t working, too. After relaunching Xcode this morning, it seems to work. I’ll go ahead and make that change now.

Sources/Algorithms/Combinations.swift
```diff
@@ -129,10 +184,76 @@ extension Combinations: Equatable where Base: Equatable {}
...
+  /// the indices in each combination are in ascending lexicographical order.
+  /// That is, in the example above, the combinations in order are the elements
+  /// at `[0]`, `[1]`, `[2]`, `[3]`, `[0, 1]`, `[0, 2]`, `[0, 3]`, `[1, 2]`,
+  /// `[1, 3]`, and finally `[2, 3]`.
+  ///
+  /// This example prints _all_ the combinations (including an empty array and
+  /// the original collection) from an array of numbers:
+  ///
+  ///     let numbers = [10, 20, 30, 40]
+  ///     for combo in numbers.combinations() {
```
kylemacomber, line 222:
```suggestion
  ///     for combo in numbers.combinations(0...) {
```

Sources/Algorithms/Combinations.swift
```diff
@@ -153,9 +274,9 @@ extension Collection {
   /// That is, in the example above, the combinations in order are the elements
   /// at `[0, 1, 2]`, `[0, 1, 3]`, `[0, 2, 3]`, and finally `[1, 2, 3]`.
   ///
-  /// If `k` is zero, the resulting sequence has exactly one element, an empty
-  /// array. If `k` is greater than the number of elements in this sequence,
-  /// the resulting sequence has no elements.
+  /// If `kRange` is zero, the resulting sequence has exactly one element, an
+  /// empty array. If `kRange` is greater than the number of elements in this
+  /// sequence, the resulting sequence has no elements.
```
kylemacomber, line 279: I think these changes are the victim of an overaggressive find/replace 😄

kylemacomber:
Does it make sense to add the support for ranges to `permutations(ofCount:)` as well? I'm not sure it needs to be implemented as a part of this PR, if it's a good idea, we could also just open an issue to track it.

Sources/Algorithms/Combinations.swift
mdznr: You’re right! It should be fixed now.

mdznr:
> Does it make sense to add the support for ranges to `permutations(ofCount:)` as well? I'm not sure it needs to be implemented as a part of this PR, if it's a good idea, we could also just open an issue to track it.

I think so. It’ll definitely produce quite a lot of permutations to iterate through, but should be done for parity. If the high-level changes for this PR are considered approved, I’ll begin work on making the changes there, too.

Sources/Algorithms/Combinations.swift
```diff
@@ -14,17 +14,54 @@ public struct Combinations<Base: Collection> {
...
-  internal var k: Int
+  internal let kRange: Range<Int>?
   
+  /// Initializes a `Combinations` for all combinations of `base` of all sizes.
+  /// - Parameter base: The collection to iterate over for combinations.
+  @usableFromInline
+  internal init(_ base: Base) {
+    self.init(base, kRange: 0...)
+  }
+
```
kylemacomber, line 29:
```suggestion  
```

I don't think this is being used anymore

mdznr: Thanks! I’ve now removed it from the original commit making the change to remove `combinations()`.

kylemacomber (approved):
I didn't go over the code and comments with a fine toothed comb, Nate's more qualified to do that, but at a more superficial level this all looks fantastic! Wonderful work @mdznr!

Sources/Algorithms/Combinations.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Algorithms open source project
 //
-// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
```
natecook1000, line 5: This should be a range now:
```suggestion
// Copyright (c) 2020-2021 Apple Inc. and the Swift project authors
```

Tests/SwiftAlgorithmsTests/CombinationsTests.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Algorithms open source project
 //
-// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
```
natecook1000, line 5:
```suggestion
// Copyright (c) 2020-2021 Apple Inc. and the Swift project authors
```

Sources/Algorithms/Combinations.swift
natecook1000: Maybe GH is doing something funky, but it doesn't look like this change got applied — should be:
```suggestion
  ///     for combo in numbers.combinations(ofCount: 0...) {
```

Sources/Algorithms/Combinations.swift
```diff
@@ -129,10 +183,76 @@ extension Combinations: Equatable where Base: Equatable {}
...
+  ///     // [30, 40]
+  ///     // [10, 20, 30]
+  ///     // [10, 20, 40]
+  ///     // [10, 30, 40]
+  ///     // [20, 30, 40]
+  ///     // [10, 20, 30, 40]
+  ///
+  /// If `kRange` is `0...0`, the resulting sequence has exactly one element, an
+  /// empty array. If `k.upperBound` is greater than the number of elements in
+  /// this sequence, the resulting sequence has no elements.
```
natecook1000, line 243: The second sentence of this paragraph doesn't seem quite right. Maybe replace it with a sentence describing how `kRange` gets limited to between zero and the length of the collection, and a note at the end saying that if that clamped range is empty, the resulting sequence has no elements?

Sources/Algorithms/Combinations.swift
```diff
@@ -129,10 +183,76 @@ extension Combinations: Equatable where Base: Equatable {}
...
+  ///
+  /// - Parameter kRange: The range of numbers of elements to include in each
+  /// combination.
+  ///
+  /// - Complexity: O(1)
```
natecook1000, line 248: This method (along with the existing one below) needs the note about O(n) for non-random-access collections in the guide. Can you add that?

natecook1000 (changes requested):
Looking close, @mdznr! 👏

Sources/Algorithms/Combinations.swift
mdznr: That’s a good idea. I’ll make that change.

Sources/Algorithms/Combinations.swift
mdznr: Good idea. I’ve made this change.

mdznr:
Thanks for the feedback, @kylemacomber and @natecook1000 . I think I’ve addressed all the feedback so far. Once this is approved/merged, I’ll create the issues (enhancements) to [optimize the performance of `counts`](https://github.com/apple/swift-algorithms/pull/51/files#r552056051) and [add parity to `Permutations`](https://github.com/apple/swift-algorithms/pull/51#issuecomment-758148669).

natecook1000 (approved):
:shipit: 🎉

Sources/Algorithms/Combinations.swift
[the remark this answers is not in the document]
mdznr: https://github.com/apple/swift-algorithms/issues/58

PR #52, opened 2020-12-22
Copy

CTMacUser:
### Description

This change adds methods that introduces new elements to a collection via overwriting existing elements (instead of inserting more space).

### Detailed Design

The are methods that copy across subsequences of the same collection, both in the forward and backward directions.  The main versions use a separate object as the source.  There are combinations for `Sequence` vs. `Collection` source, and copying over the prefix vs. suffix.  There is another main version that copies suffix to suffix.

```swift
extension MutableCollection {
    /// Copies the prefix of the given sequence on top of the prefix of this collection.
    public mutating func copy<S>(from source: S) -> (copyEnd: Index, sourceTail: S.Iterator) where S : Sequence, Self.Element == S.Element

    /// Copies the prefix of the given collection on top of the prefix of this collection.
    public mutating func copy<C>(collection: C) -> (copyEnd: Index, sourceTailStart: C.Index) where C : Collection, Self.Element == C.Element

    /// Copies, in forward traversal, the prefix of a subsequence on top of the prefix of another, using the given bounds to demarcate the subsequences.
    public mutating func copy<R, S>(forwardsFrom source: R, to destination: S) -> (sourceRead: Range<Index>, destinationWritten: Range<Index>) where R : RangeExpression, S : RangeExpression, Self.Index == R.Bound, R.Bound == S.Bound
}

extension MutableCollection where Self : BidirectionalCollection {
    /// Copies the prefix of the given sequence on top of the suffix of this collection.
    public mutating func copy<S>(asSuffix source: S) -> (copyStart: Index, sourceTail: S.Iterator) where S : Sequence, Self.Element == S.Element

    /// Copies the prefix of the given collection on top of the suffix of this collection.
    public mutating func copy<C>(collectionAsSuffix source: C) -> (copyStart: Index, sourceTailStart: C.Index) where C : Collection, Self.Element == C.Element

    /// Copies the suffix of the given collection on top of the suffix of this collection.
    public mutating func copy<C>(backwards source: C) -> (writtenStart: Index, readStart: C.Index) where C : BidirectionalCollection, Self.Element == C.Element

    /// Copies, in reverse traversal, the suffix of a subsequence on top of the suffix of another, using the given bounds to demarcate the subsequences.
    public mutating func copy<R, S>(backwardsFrom source: R, to destination: S) -> (sourceRead: Range<Index>, destinationWritten: Range<Index>) where R : RangeExpression, S : RangeExpression, Self.Index == R.Bound, R.Bound == S.Bound
}
```

### Documentation Plan

The methods have documentation comments.  There is also a guide file.

### Test Plan

There is a XC Test file for these methods.

### Source Impact

The change adds API.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@ package updates, you can specify your package dependency using
   the copying ended and an iterator for the source sequence after the elements
   that were copied.  The `copy(collection:)` method works like the previous
   method, but uses a collection as the source, and expresses the unread suffix
-  for that source as an `Index` instead.  The `copyOntoSuffix(with:)` and
-  `copyOntoSuffix(withCollection:)` methods work like the first two methods
+  for that source as an `Index` instead.  The `copy(asSuffix:)` and
+  `copy(collectionAsSuffix:)` methods work like the first two methods
   except the end of the receiver is overwritten instead of the beginning, and
   so their return value instead includes the starting index in the receiver
   where the copying began.  The `copy(backwards:)` method works like the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ package updates, you can specify your package dependency using
   where the copying began.  The `copy(backwards:)` method works like the
   previous method, except the source is also read from the end instead of the
   beginning, and so the return values are the starting indices of both
-  collections' targeted elements.
+  collections' targeted elements.  The Swift memory model restricts reading and
+  writing into the same collection, so the `copy(forwardsFrom:to:)` and
+  `copy(backwardsFrom:to:)` methods provide same-collection element copying.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,14 @@ package updates, you can specify your package dependency using
 
 ## [Unreleased]
 
-*No changes yet.*
+### Additions
+
+- The `copy(from:)` method has been added, applying to types conforming to
+  `MutableCollection`.  It takes a sequence with the same element type as its
+  only parameter, whose elements will be copied on top of the existing
+  elements.  The return values are the past-the-end index in the receiver where
+  the copying ended and an iterator for the source sequence after the elements
+  that were copied.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,11 @@ package updates, you can specify your package dependency using
   the copying ended and an iterator for the source sequence after the elements
   that were copied.  The `copy(collection:)` method works like the previous
   method, but uses a collection as the source, and expresses the unread suffix
-  for that source as an `Index` instead.
+  for that source as an `Index` instead.  The `copyOntoSuffix(with:)` and
+  `copyOntoSuffix(withCollection:)` methods work like the first two methods
+  except the end of the receiver is overwritten instead of the beginning, and
+  so their return value instead includes the starting index in the receiver
+  where the copying began.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,10 @@ package updates, you can specify your package dependency using
   `copyOntoSuffix(withCollection:)` methods work like the first two methods
   except the end of the receiver is overwritten instead of the beginning, and
   so their return value instead includes the starting index in the receiver
-  where the copying began.
+  where the copying began.  The `copy(backwards:)` method works like the
+  previous method, except the source is also read from the end instead of the
+  beginning, and so the return values are the starting indices of both
+  collections' targeted elements.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,9 @@ package updates, you can specify your package dependency using
   only parameter, whose elements will be copied on top of the existing
   elements.  The return values are the past-the-end index in the receiver where
   the copying ended and an iterator for the source sequence after the elements
-  that were copied.
+  that were copied.  The `copy(collection:)` method works like the previous
+  method, but uses a collection as the source, and expresses the unread suffix
+  for that source as an `Index` instead.
 
 ---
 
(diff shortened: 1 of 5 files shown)


mdznr (commented):
1. **Naming** 
    I wouldn’t have guessed this would be the behavior of a function named `copy`. This may be because I’ve programmed a lot more Objective-C than C++ recently. I suspect many Swift programmers are also in that same situation.

    Perhaps the naming should include some clarifying words, like “prefix”, or “over”/“overwrite”?

    `overwritePrefix(with:)`?

2. **`Sequence` and `Collection`**
    I was a bit confused as to why there is a `copy(from:)` taking a `Sequence` and a differently named `copy(collection:)` (omitting the “from” terminology) for `Collection`.
    
    If I’m working with two `Array`s, it’s not clear which function I should use and why.
    
    ```swift
    var destination = ["A", "B", "C", "D", "E", "F"]
    let source = ["a", "b", "c"]
    destination.copy(from: source)
    destination.copy(collection: source)
    ````
    
    Are these both necessary? And if so, how can it be made more clear which function to use?

3. **Return Value**
    Should these functions be marked as `@discardableResult`? Seems like it’s valid to call these functions without making use of the return value.

4. **Performance**
    There are a few shortcuts that could probably made depending on the input sizes. As one example, if the source is larger than the destination, why bother going through the items one-by-one when a prefix of `self.count` can be taken from the source?

----

**Minor:**

I’ve noticed that the rest of this repository uses a single between sentences, but these changes include comments and documentation using two. This should be consistent.

CTMacUser:
> 1. **Naming**
>    I wouldn’t have guessed this would be the behavior of a function named `copy`. This may be because I’ve programmed a lot more Objective-C than C++ recently. I suspect many Swift programmers are also in that same situation.
>
>    Perhaps the naming should include some clarifying words, like “prefix”, or “over”/“overwrite”?
>
>    `overwritePrefix(with:)`?

Due to the nature of C++'s output iterator concept, its copying includes both memory overwriting **and** new-space insertion!  Swift puts those as two separate collection concepts (`MutableCollection` vs. `RangeReplaceableCollection`).  `RRC` already covers copying that inserts new memory spaces, but they don't use "copy" in their names (`replaceSubrange`, `append`, and `insert`), so "copy" was free to use.  Maybe "overwrite" or similar would be better.

> 2. **`Sequence` and `Collection`**
>    I was a bit confused as to why there is a `copy(from:)` taking a `Sequence` and a differently named `copy(collection:)` (omitting the “from” terminology) for `Collection`.
>
>    If I’m working with two `Array`s, it’s not clear which function I should use and why.
>
>    ```swift
>    var destination = ["A", "B", "C", "D", "E", "F"]
>    let source = ["a", "b", "c"]
>    destination.copy(from: source)
>    destination.copy(collection: source)
>    ```
>    
>    Are these both necessary? And if so, how can it be made more clear which function to use?

The sequence vs. collection methods differ in their return type.  Since a `Sequence` may be single-pass, I can only return an `Iterator` to the unused elements.  For a `Collection`, I return an `Index` to the first unused element (or `endIndex` if all were used).  The interfaces for the second member of the return value are obviously incompatible, so I can't/shouldn't reuse a (full) name for both.

Hmm, would a better option than a `Sequence` be an `inout IteratorProtocol` parameter instead?  But that would put the extra step of converting onto the user.  (A similar thing happened when I proposed an initializer to `CollectionOfOne` that reads the first element from a sequence/iterator.  Maybe the same solution will work here.)

> 3. **Return Value**
>    Should these functions be marked as `@discardableResult`? Seems like it’s valid to call these functions without making use of the return value.

I wonder about that, too.  Since it's (currently) non-discardable, the user has to think about what if one of the collections/sequences wasn't completely used.  If the user knows that the source and destination are the same length, or otherwise doesn't care, they can use the "`_ =`" syntax.

> 4. **Performance**
>    There are a few shortcuts that could probably made depending on the input sizes. As one example, if the source is larger than the destination, why bother going through the items one-by-one when a prefix of `self.count` can be taken from the source?

Are you implying the existence of a procedure that can do an O(1) bulk overwriting-copy?  Otherwise, this wouldn't save any time.  And if there is a (refined) protocol that can guarantee this (and also conform to `RandomAccessCollection` so `self.count` would be O(1)), that wouldn't help with any other sequence.

> **Minor:**
> 
> I’ve noticed that the rest of this repository uses a single between sentences, but these changes include comments and documentation using two. This should be consistent.

I'll check that out.

CTMacUser:
I renamed the methods from "copy" to "overwrite" and did a lot of the other changes suggested by @mdznr.  Check it out.

Guides/Overwrite.md
```diff
@@ -0,0 +1,124 @@
...
+`overwrite(suffixUsing:)` methods use their source's prefix, while the
+`overwrite(backwardsWith:)` and `overwrite(backwardsFrom:to:)` methods use
+their source's suffix.
+
+## Detailed Design
+
+New methods are added to element-mutable collections:
+
+```swift
+extension MutableCollection {
```
LucianoPAlmeida, line 39: This indeed seems a nice proposal and well implemented, but I was wondering that this functionality is kind of already covered in a more general way by `RangeReplaceableCollection`'s `replaceSubrange(:with:)`. 
So those are more questions/suggestions: 
What are the advantages of having this API over the existing one? 
And also, related to naming, could this complement the existing one by being called `replacePrefix(with:)`? 
I think this is more align with the convention and easier to reason about given the existing API's available.

Guides/Overwrite.md
```diff
@@ -0,0 +1,124 @@
...
+
+Calling these methods is O(_k_), where _k_ is the length of the shorter
+(virtual) sequence between the receiver (subsequence) and the source.
+
+### Naming
+
+The initial development version of this library used the term-of-art "`copy`"
+as the base name of this family of methods. But since the insertion-copying
+methods (in `RangeReplaceableCollection`) do not use the term, and the term is
+used for object copying in Foundation, a subsitute term was chosen here. The
```
LucianoPAlmeida, line 102: Typo?
```suggestion
used for object copying in Foundation, a substitute term was chosen here. The
```

Guides/Overwrite.md
CTMacUser: `RangeReplaceableCollection` and `MutableCollection` generally can’t be substituted for each other; the API you mentioned is pretty much their sole spiritual overlap, and it wouldn’t work with collections that implement only `MC`.

adincebic:
@swift-ci Please test

PR #54, opened 2020-12-24
[Chunked] Add `chunks(of:)` variant that divides a collection in chunks of a given size

LucianoPAlmeida:
This just adds the `chunks(of:)` to chunk a collection into subsequences of a given size. 
As suggested by @bjhomer [here](https://forums.swift.org/t/pull-in-chunks-ofsize-from-separate-proposal/42575) this is just moving a long time opened [evolution proposal](https://forums.swift.org/t/pull-in-chunks-ofsize-from-separate-proposal/42575) to this package. So let me know what you think =] 

cc @natecook1000 @kylemacomber 

### Checklist

### DIFF
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -392,15 +392,13 @@ extension ChunkedByCount {
     guard limit != i else { return nil }
     
     if offset > 0 {
-      guard limit < i || distance(from: i, to: limit) >= offset else {
-        return nil
-      }
-      return offsetForward(i, offsetBy: offset)
+      return limit > i
+        ? offsetForward(i, offsetBy: offset, limit: limit)
+        : offsetForward(i, offsetBy: offset)
     } else {
-      guard limit > i || distance(from: i, to: limit) <= offset else {
-        return nil
-      }
-      return offsetBackward(i, offsetBy: offset)
+      return limit < i
+        ? offsetBackward(i, offsetBy: offset, limit: limit)
+        : offsetBackward(i, offsetBy: offset)
     }
   }
 
@@ -408,50 +406,98 @@ extension ChunkedByCount {
   public func index(_ i: Index, offsetBy distance: Int) -> Index {
     guard distance != 0 else { return i }
     
-    return distance > 0
+    let idx = distance > 0
         ? offsetForward(i, offsetBy: distance)
         : offsetBackward(i, offsetBy: distance)
+    guard let index = idx else {
+      fatalError("Out of bounds")
+    }
+    return index
   }
   
   @usableFromInline
-  internal func offsetForward(_ i: Index, offsetBy distance: Int) -> Index {
+  internal func offsetForward(
+    _ i: Index, offsetBy distance: Int, limit: Index? = nil
+  ) -> Index? {
     assert(distance > 0)
+
     return makeOffsetIndex(
-      from: i, baseBound: base.endIndex, baseDistance: distance * chunkCount
+      from: i, baseBound: base.endIndex,
+      distance: distance, baseDistance: distance * chunkCount,
+      limit: limit, by: >
     )
   }
   
-  @usableFromInline
-  internal func offsetBackward(_ i: Index, offsetBy distance: Int) -> Index {
-    assert(distance < 0)
-    if i.baseRange.lowerBound == base.endIndex {
+  // Convenience to compute offset backward base distance.
+  @inline(__always)
+  private func computeOffsetBackwardBaseDistance(
+    _ i: Index, _ distance: Int
+  ) -> Int {
+    if i == endIndex {
       let remainder = base.count%chunkCount
       // We have to take it into account when calculating offsets.
       if remainder != 0 {
-        // Distance "minus" one(at this point distance is negative) because we
-        // need to adjust for the last position that have a variadic(remainder)
-        // number of elements.
-        let baseDistance = ((distance + 1) * chunkCount) - remainder
-        return makeOffsetIndex(
-          from: i, baseBound: base.startIndex, baseDistance: baseDistance
-        )
+        // Distance "minus" one(at this point distance is negative)
+        // because we need to adjust for the last position that have
+        // a variadic(remainder) number of elements.
+        return ((distance + 1) * chunkCount) - remainder
       }
     }
+    return distance * chunkCount
+  }
+  
+  @usableFromInline
+  internal func offsetBackward(
+    _ i: Index, offsetBy distance: Int, limit: Index? = nil
+  ) -> Index? {
+    assert(distance < 0)
+    let baseDistance =
+        computeOffsetBackwardBaseDistance(i, distance)
     return makeOffsetIndex(
-      from: i, baseBound: base.startIndex, baseDistance: distance * chunkCount
+      from: i, baseBound: base.startIndex,
+      distance: distance, baseDistance: baseDistance,
+      limit: limit, by: <
     )
   }
   
   // Helper to compute index(offsetBy:) index.
   @inline(__always)
   private func makeOffsetIndex(
-    from i: Index, baseBound: Base.Index, baseDistance: Int
-  ) -> Index {
-    let baseStartIdx = base.index(
+    from i: Index, baseBound: Base.Index, distance: Int, baseDistance: Int,
+    limit: Index?, by limitFn: (Base.Index, Base.Index) -> Bool
+  ) -> Index? {
+    let baseIdx = base.index(
       i.baseRange.lowerBound, offsetBy: baseDistance,
       limitedBy: baseBound
-    ) ?? baseBound
+    )
     
+    if let limit = limit {
+      if baseIdx == nil {
+        // If we past the bounds while advancing forward and the
+        // limit is the `endIndex`, since the computation on base
+        // don't take into account the remainder, we have to make
+        // sure that passing the bound was because of the distance
+        // not just because of a remainder. Special casing is less
+        // expensive than always use count(which could be O(n) for
+        // non-random access collection base) to compute the base
+        // distance taking remainder into account.
+        if baseDistance > 0 && limit == endIndex {
+          if self.distance(from: i, to: limit) < distance {
+            return nil
+          }
+        } else {
+          return nil
+        }
+      }
+
+      // Checks for the limit.
+      let baseStartIdx = baseIdx ?? baseBound
+      if limitFn(baseStartIdx, limit.baseRange.lowerBound) {
+        return nil
+      }
+    }
+    
+    let baseStartIdx = baseIdx ?? baseBound
     let baseEndIdx = base.index(
       baseStartIdx, offsetBy: chunkCount, limitedBy: base.endIndex
     ) ?? base.endIndex
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -313,11 +313,13 @@ extension ChunkedByCount: Collection {
   
   /// - Complexity: O(n)
   public subscript(i: Index) -> Element {
-    base[i.baseRange]
+    precondition(i < endIndex, "Index out of range")
+    return base[i.baseRange]
   }
   
   @inlinable
   public func index(after i: Index) -> Index {
+    precondition(i < endIndex, "Advancing past end index")
     let baseIdx = base.index(
       i.baseRange.upperBound, offsetBy: chunkCount,
       limitedBy: base.endIndex
@@ -339,6 +341,8 @@ extension ChunkedByCount:
 where Base: RandomAccessCollection {
   @inlinable
   public func index(before i: Index) -> Index {
+    precondition(i > startIndex, "Advancing past start index")
+    
     var offset = chunkCount
     if i.baseRange.lowerBound == base.endIndex {
       let remainder = base.count%chunkCount
@@ -354,6 +358,10 @@ where Base: RandomAccessCollection {
     return Index(_baseRange: baseIdx..<i.baseRange.lowerBound)
   }
   
+  // TODO: index(_:offsetBy:) and index(_:offsetBy:limitedBy:)
+}
+
+extension ChunkedByCount {
   @inlinable
   public func distance(from start: Index, to end: Index) -> Int {
     let distance =
@@ -361,18 +369,14 @@ where Base: RandomAccessCollection {
                     to: end.baseRange.lowerBound)
     let (quotient, remainder) =
       distance.quotientAndRemainder(dividingBy: chunkCount)
-    // Increment should account for negative distances.
-    if remainder < 0 {
-      return quotient - 1
-    }
-    return quotient + (remainder == 0 ? 0 : 1)
+    return quotient + remainder.signum()
   }
 
   @inlinable
   public var count: Int {
     let (quotient, remainder) =
       base.count.quotientAndRemainder(dividingBy: chunkCount)
-    return quotient + (remainder == 0 ? 0 : 1)
+    return quotient + remainder.signum()
   }
 }
 
@@ -404,10 +408,10 @@ extension Collection {
 extension ChunkedByCount: Equatable where Base: Equatable {}
 
 // Since we have another stored property of type `Index` on the
-// collection, synthetization of hashble conformace would require
+// collection, synthesis of `Hashble` conformace would require
 // a `Base.Index: Hashable` constraint, so we implement the hasher
-// only in terms of base. Since the computed index is based on it,
-// it should make a difference here.
+// only in terms of `base`. Since the computed index is based on it,
+// it should not make a difference here.
 extension ChunkedByCount: Hashable where Base: Hashable {
   public func hash(into hasher: inout Hasher) {
     hasher.combine(base)
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -305,13 +305,15 @@ extension ChunkedByCount: Collection {
     }
   }
 
-  /// - Complexity: O(n)
+  /// - Complexity: O(1)
+  @inlinable
   public var startIndex: Index { computedStartIndex }
+  @inlinable
   public var endIndex: Index {
     Index(_baseRange: base.endIndex..<base.endIndex)
   }
   
-  /// - Complexity: O(n)
+  /// - Complexity: O(1)
   public subscript(i: Index) -> Element {
     precondition(i < endIndex, "Index out of range")
     return base[i.baseRange]
@@ -358,7 +360,81 @@ where Base: RandomAccessCollection {
     return Index(_baseRange: baseIdx..<i.baseRange.lowerBound)
   }
   
-  // TODO: index(_:offsetBy:) and index(_:offsetBy:limitedBy:)
+  @inlinable
+  public func index(
+    _ i: Index, offsetBy offset: Int, limitedBy limit: Index
+  ) -> Index? {
+    guard offset != 0 else { return i }
+    guard limit != i else { return nil }
+    
+    if offset > 0 {
+      guard limit <= i || distance(from: i, to: limit) >= offset else {
+        return nil
+      }
+      return offsetForward(i, offsetBy: offset)
+    } else {
+      guard limit >= i || distance(from: i, to: limit) <= offset else {
+        return nil
+      }
+      return offsetBackward(i, offsetBy: offset)
+    }
+  }
+
+  @inlinable
+  public func index(_ i: Index, offsetBy distance: Int) -> Index {
+    guard distance != 0 else { return i }
+    
+    return distance > 0
+        ? offsetForward(i, offsetBy: distance)
+        : offsetBackward(i, offsetBy: distance)
+  }
+  
+  @usableFromInline
+  internal func offsetForward(_ i: Index, offsetBy distance: Int) -> Index {
+    return makeOffsetIndex(
+      from: i, baseBound: base.endIndex, distance: distance
+    )
+  }
+  
+  @usableFromInline
+  internal func offsetBackward(_ i: Index, offsetBy n: Int) -> Index {
+    var idx = i
+    var distance = n
+    // If we know that the last chunk is the only one that can possible
+    // have a variadic count. So in order to simplify and avoid another
+    // calculation of offsets(that is already done at `index(before:)`)
+    // we just move one position already so the index can be calculated
+    // assuming all chunks have the same size.
+    if i.baseRange.lowerBound == base.endIndex {
+      formIndex(before: &idx)
+      distance += 1
+      // If the offset was simply one, we are done.
+      guard distance != 0 else {
+        return idx
+      }
+    }
+
+    return makeOffsetIndex(
+      from: idx, baseBound: base.startIndex, distance: distance
+    )
+  }
+  
+  // Helper to compute index(offsetBy:) index.
+  internal func makeOffsetIndex(
+    from i: Index, baseBound: Base.Index, distance: Int
+  ) -> Index {
+    let baseStartIdx = base.index(
+      i.baseRange.lowerBound, offsetBy: distance * chunkCount,
+      limitedBy: baseBound
+    ) ?? baseBound
+    
+    let baseEndIdx = base.index(
+      i.baseRange.lowerBound, offsetBy: (distance + 1) * chunkCount,
+      limitedBy: base.endIndex
+    ) ?? base.endIndex
+    
+    return Index(_baseRange: baseStartIdx..<baseEndIdx)
+  }
 }
 
 extension ChunkedByCount {
@@ -399,7 +475,7 @@ extension Collection {
   /// - Complexity: O(1)
   @inlinable
   public func chunks(ofCount count: Int) -> ChunkedByCount<Self> {
-    precondition(count > 0, " Cannot chunk with count <= 0!")
+    precondition(count > 0, "Cannot chunk with count <= 0!")
     return ChunkedByCount(_base: self, _chunkCount: count)
   }
 }
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -246,3 +246,177 @@ extension Collection {
     try chunked(on: projection, by: ==)
   }
 }
+
+//===----------------------------------------------------------------------===//
+// chunks(ofCount:)
+//===----------------------------------------------------------------------===//
+
+/// A collection that presents the elements of its base collection
+/// in `SubSequence` chunks of any given count.
+///
+/// A `ChunkedByCount` is a lazy view on the base Collection, but it does not implicitly confer
+/// laziness on algorithms applied to its result.  In other words, for ordinary collections `c`:
+///
+/// * `c.chunks(ofCount: 3)` does not create new storage
+/// * `c.chunks(ofCount: 3).map(f)` maps eagerly and returns a new array
+/// * `c.lazy.chunks(ofCount: 3).map(f)` maps lazily and returns a `LazyMapCollection`
+public struct ChunkedByCount<Base: Collection> {
+  
+  public typealias Element = Base.SubSequence
+  
+  @usableFromInline
+  internal let base: Base
+  
+  @usableFromInline
+  internal let chunkCount: Int
+  
+  @usableFromInline
+  internal var computedStartIndex: Index
+
+  ///  Creates a view instance that presents the elements of `base`
+  ///  in `SubSequence` chunks of the given count.
+  ///
+  /// - Complexity: O(n)
+  @inlinable
+  internal init(_base: Base, _chunkCount: Int) {
+    self.base = _base
+    self.chunkCount = _chunkCount
+    
+    // Compute the start index upfront in order to make
+    // start index a O(1) lookup.
+    let baseEnd = _base.index(
+      _base.startIndex, offsetBy: _chunkCount,
+      limitedBy: _base.endIndex
+    ) ?? _base.endIndex
+    
+    self.computedStartIndex =
+      Index(_baseRange: _base.startIndex..<baseEnd)
+  }
+}
+
+extension ChunkedByCount: Collection {
+  public struct Index {
+    @usableFromInline
+    internal let baseRange: Range<Base.Index>
+    
+    @usableFromInline
+    internal init(_baseRange: Range<Base.Index>) {
+      self.baseRange = _baseRange
+    }
+  }
+
+  /// - Complexity: O(n)
+  public var startIndex: Index { computedStartIndex }
+  public var endIndex: Index {
+    Index(_baseRange: base.endIndex..<base.endIndex)
+  }
+  
+  /// - Complexity: O(n)
+  public subscript(i: Index) -> Element {
+    base[i.baseRange]
+  }
+  
+  @inlinable
+  public func index(after i: Index) -> Index {
+    let baseIdx = base.index(
+      i.baseRange.upperBound, offsetBy: chunkCount,
+      limitedBy: base.endIndex
+    ) ?? base.endIndex
+    return Index(_baseRange: i.baseRange.upperBound..<baseIdx)
+  }
+}
+
+extension ChunkedByCount.Index: Comparable {
+  @inlinable
+  public static func < (lhs: ChunkedByCount.Index,
+                        rhs: ChunkedByCount.Index) -> Bool {
+    lhs.baseRange.lowerBound < rhs.baseRange.lowerBound
+  }
+}
+
+extension ChunkedByCount:
+  BidirectionalCollection, RandomAccessCollection
+where Base: RandomAccessCollection {
+  @inlinable
+  public func index(before i: Index) -> Index {
+    var offset = chunkCount
+    if i.baseRange.lowerBound == base.endIndex {
+      let remainder = base.count%chunkCount
+      if remainder != 0 {
+        offset = remainder
+      }
+    }
+    
+    let baseIdx = base.index(
+      i.baseRange.lowerBound, offsetBy: -offset,
+      limitedBy: base.startIndex
+    ) ?? base.startIndex
+    return Index(_baseRange: baseIdx..<i.baseRange.lowerBound)
+  }
+  
+  @inlinable
+  public func distance(from start: Index, to end: Index) -> Int {
+    let distance =
+      base.distance(from: start.baseRange.lowerBound,
+                    to: end.baseRange.lowerBound)
+    let (quotient, remainder) =
+      distance.quotientAndRemainder(dividingBy: chunkCount)
+    // Increment should account for negative distances.
+    if remainder < 0 {
+      return quotient - 1
+    }
+    return quotient + (remainder == 0 ? 0 : 1)
+  }
+
+  @inlinable
+  public var count: Int {
+    let (quotient, remainder) =
+      base.count.quotientAndRemainder(dividingBy: chunkCount)
+    return quotient + (remainder == 0 ? 0 : 1)
+  }
+}
+
+extension Collection {
+  /// Returns a `ChunkedCollection<Self>` view presenting the elements
+  /// in chunks with count of the given count parameter.
+  ///
+  /// - Parameter size: The size of the chunks. If the count parameter
+  ///   is evenly divided by the count of the base `Collection` all the
+  ///   chunks will have the count equals to size.
+  ///   Otherwise, the last chunk will contain the remaining elements.
+  ///
+  ///     let c = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+  ///     print(c.chunks(ofCount: 5).map(Array.init))
+  ///     // [[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]]
+  ///
+  ///     print(c.chunks(ofCount: 3).map(Array.init))
+  ///     // [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10]]
+  ///
+  /// - Complexity: O(1)
+  @inlinable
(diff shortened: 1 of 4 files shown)


kylemacomber:
I think this would make a great addition!

What do you think about the name `chunks(ofCount:)` similar to [`combinations(ofCount:)`](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Combinations.swift#L164)? I think this would make it clearer that the argument isn't specifying an element being searched for (like `firstIndex(of:)`):

```swift
let x = [1, 2, 2, 3, 3, 3, 2]
print(x.chunks(of: 2))
// [[2, 2], [2]] or [[1, 2], [2, 3], [3, 3], [2]]?
```

But rather the chunk size:

```swift
let x = [1, 2, 2, 3, 3, 3, 2]
print(x.chunks(ofCount: 2))
// [[1, 2], [2, 3], [3, 3], [2]]
```

LucianoPAlmeida:
Hey @kylemacomber :)

> What do you think about the name chunks(ofCount:) similar to combinations(ofCount:)? I think this would make it clearer that the argument isn't specifying an element being searched for (like firstIndex(of:)):

I agree, that is a good idea and will remove any ambiguity as far as interpreting the function... I'll make the changes :)

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,136 @@ extension Collection {
...
+    }
+    return Index(_base: base.index(i.base, offsetBy: -_chunkCount))
+  }
+  
+  @inlinable
+  public func distance(from start: Index, to end: Index) -> Int {
+    let distance = base.distance(from: start.base, to: end.base)
+    let (quotient, remainder) =
+      distance.quotientAndRemainder(dividingBy: _chunkCount)
+    return quotient + remainder
```
kylemacomber, line 334: I'm not sure this is correct, for example, with a collection of count 5 and chunk size of 3. That should be 2 chunks, not 3.
```suggestion
    return quotient + (remainder == 0 ? 0 : 1)
```

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,136 @@ extension Collection {
...
+  ///     print(c.chunks(ofCount: 5).map(Array.init))
+  ///     // [[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]]
+  ///
+  ///     print(c.chunks(ofCount: 3).map(Array.init))
+  ///     // [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10]]
+  ///
+  /// - Complexity: O(1)
+  @inlinable
+  public func chunks(ofCount count: Int) -> ChunkedCollection<Self> {
+    assert(count > 0, " Cannot chunk with size <= 0!")
```
kylemacomber, line 374: Maybe this should be a precondition, so it also traps in release builds?
```suggestion
    precondition(count > 0, " Cannot chunk with count <= 0!")
```

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,136 @@ extension Collection {
...
+    return Index(_base:
+      base.index(i.base, offsetBy: n * _chunkCount, limitedBy: bound) ?? bound
+    )
+  }
+  
+  @inlinable
+  public var count: Int {
+    return base.count
+      .isMultiple(of: _chunkCount) ? base.count/_chunkCount
+                                   : base.count/_chunkCount + 1
```
kylemacomber, line 351: `count` and `distance` seem to be doing the same calculation, just expressed with different code. If we're going to repeat ourselves, we should at least use the same code, though it's probably best to try to avoid repeating ourselves

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Indeed, thanks for catching this :)

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Ah I just noted that default implementation of collection `count` is in terms of distance, so I think we don't even need this

LucianoPAlmeida:
All fixed Kyle =]

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,136 @@ extension Collection {
...
+
+/// A collection that presents the elements of its base collection
+/// in `SubSequence` chunks of any given size.
+///
+/// A ChunkedCollection is a lazy view on the base Collection, but it does not implicitly confer
+/// laziness on algorithms applied to its result.  In other words, for ordinary collections `c`:
+///
+/// * `c.chunks(of: 3)` does not create new storage
+/// * `c.chunks(of: 3).map(f)` maps eagerly and returns a new array
+/// * `c.lazy.chunks(of: 3).map(f)` maps lazily and returns a `LazyMapCollection`
```
natecook1000, line 262: For this to be true you’ll need to add conditional conformance to `LazyCollectionProtocol` when the base collection conforms.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,136 @@ extension Collection {
...
+public struct ChunkedCollection<Base: Collection> {
```
natecook1000, line 263: I’d like this name and the other chunking collection to work together a little better. As is, they would be `LazyChunked` and `ChunkedCollection`, which doesn’t really distinguish either. It’s a bit easier to add the semantics of this one to the name — what do you think about `struct ChunkedByCount<Base: ...>`?

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,133 @@ extension Collection {
...
+  
+  @inlinable
+  public func index(_ i: Index, offsetBy n: Int) -> Index {
+    guard n != 0 else { return i }
+    
+    let bound = n > 0 ? base.endIndex : base.startIndex
+    return Index(_base:
+      base.index(i.base, offsetBy: n * _chunkCount, limitedBy: bound) ?? bound
+    )
+  }
```
natecook1000, line 349: These should both move up into an unconstrained extension.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,133 @@ extension Collection {
...
+  @inlinable
+  public func chunks(ofCount count: Int) -> ChunkedCollection<Self> {
+    precondition(count > 0, " Cannot chunk with count <= 0!")
+    return ChunkedCollection(_base: self, _chunkCount: count)
+  }
+}
+
+// Conditional conformances.
+extension ChunkedCollection: Equatable where Base: Equatable {}
+extension ChunkedCollection: Hashable where Base: Hashable {}
```
natecook1000, line 378: Let’s add `Hashable` conformance for the index, too.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,133 @@ extension Collection {
...
+  public subscript(i: Index) -> Element {
+    base[i.base..<index(after: i).base]
+  }
+  
+  @inlinable
+  public func index(after i: Index) -> Index {
+    let baseIdx = base.index(
+      i.base, offsetBy: _chunkCount, limitedBy: base.endIndex)
+    return Index(_base: baseIdx ?? base.endIndex)
+  }
```
natecook1000, line 304: Let’s add a customized version of `count` that depends on `base.count`, as well.

Tests/SwiftAlgorithmsTests/ChunkedTests.swift
```diff
@@ -76,4 +76,105 @@ final class ChunkedTests: XCTestCase {
...
+    XCTAssertEqual([Int]().chunks(ofCount: 5).count, 0)
+
+    let collection = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+    XCTAssertEqual(collection.chunks(ofCount: 1).count, 10)
+    XCTAssertEqual(collection.chunks(ofCount: 3).count, 4)
+    XCTAssertEqual(collection.chunks(ofCount: 5).count, 2)
+    XCTAssertEqual(collection.chunks(ofCount: 11).count, 1)
+  }
+  
+  func testChunksOfSizeDistance() {
```
natecook1000, line 123: You should be able to replace these last two tests with a call to `validateIndexTraversals`.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,133 @@ extension Collection {
...
+    return quotient + (remainder == 0 ? 0 : 1)
+  }
+  
+  @inlinable
+  public func index(_ i: Index, offsetBy n: Int) -> Index {
+    guard n != 0 else { return i }
+    
+    let bound = n > 0 ? base.endIndex : base.startIndex
+    return Index(_base:
+      base.index(i.base, offsetBy: n * _chunkCount, limitedBy: bound) ?? bound
```
natecook1000, line 347: Instead of landing at `bound` if it goes past, this method should trap if `base.index(_:offsetBy:limitedBy:)` returns `nil`.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,133 @@ extension Collection {
...
+  @inlinable
+  internal init(_base: Base, _chunkCount: Int) {
+    self.base = _base
+    self._chunkCount = _chunkCount
+  }
+}
+
+extension ChunkedCollection: Collection {
+  public struct Index {
+    public let base: Base.Index
```
natecook1000, line 284: This should also be internal.

There’s a bit of decision to make here between using this single index and using the range for the chunk. For a non-random-access collection, a single index makes using the subscript O(n), while a range moves that calculation to `startIndex`. Given that `startIndex` is so frequently touched for testing boundary conditions, the current implementation is probably the right one, but I don’t like that iterating over a collection this way requires traversing the base collection twice (once for each subscript access and once for each call to `index(after:)`).

Another approach would be to calculate `startIndex` at initialization and store it along with the chunk size. That way both `startIndex` and subscripting would always be O(1). What do you think?

natecook1000 (changes requested):
Happy to see this PR, @LucianoPAlmeida! I’ve done an initial review pass over the implementation — let me know if you have any questions.

Sources/Algorithms/Chunked.swift
natecook1000: Since `count` is a customization point for `Collection`, it’s good to implement it in wrapper collections that can directly calculate the new count.

LucianoPAlmeida:
Thanks for the review @natecook1000!

Sources/Algorithms/Chunked.swift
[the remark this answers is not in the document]
LucianoPAlmeida: This was fixed, as @kylemacomber pointed it too :)

Sources/Algorithms/Chunked.swift
kylemacomber: IIUC you're suggesting we store a Range in the index and calculate it in `init`, which would allow both `startIndex` and `subscript` to be O(1)? I like that idea!

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: I like it, it takes away the possible ambiguity

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Conformance added

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Ah ok, so I'll add it back :)

LucianoPAlmeida: Added!

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: > Another approach would be to calculate `startIndex` at initialization and store it along with the chunk size. That way both `startIndex` and subscripting would always be O(1). What do you think?

I've experimented with this approach when putting up this PR, I like the idea ... but there are two points:
* Both start index and endIndex(lower bound) will have to be computed. 
* If we compute at initialization(start and end indexes) that the call then `[1,2,3].chunks(ofCount: 2)` becomes O(n) in non-`RandomAccessCollection` base.

LucianoPAlmeida: Just changed :)

Tests/SwiftAlgorithmsTests/ChunkedTests.swift
LucianoPAlmeida: Done!

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: After implemented, I would say that I definitely like this approach!

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,177 @@ extension Collection {
...
+    
+    let baseIdx = base.index(
+      i.baseRange.lowerBound, offsetBy: -offset,
+      limitedBy: base.startIndex
+    ) ?? base.startIndex
+    return Index(_baseRange: baseIdx..<i.baseRange.lowerBound)
+  }
+  
+  @inlinable
+  public func distance(from start: Index, to end: Index) -> Int {
```
timvermeulen, line 358: `distance(from:to:)` and `count` are both part of `Collection` rather than `BidirectionalCollection` or `RandomAccesCollection`, so these implementations should live in the `: Collection` instead in order for collections that aren't random-access to also benefit from any potential optimizations `bas.distance(from:to:)` / `base.count` make.

And related: `RandomAccessCollection` also needs an O(1) `index(_:offsetBy:limitedBy:)` implementation, so a TODO to implement this later would be greatly appreciated 🙂

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,177 @@ extension Collection {
...
+    let distance =
+      base.distance(from: start.baseRange.lowerBound,
+                    to: end.baseRange.lowerBound)
+    let (quotient, remainder) =
+      distance.quotientAndRemainder(dividingBy: chunkCount)
+    // Increment should account for negative distances.
+    if remainder < 0 {
+      return quotient - 1
+    }
+    return quotient + (remainder == 0 ? 0 : 1)
```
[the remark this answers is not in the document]
kylemacomber: TIL

Sources/Algorithms/Chunked.swift
natecook1000: Great, thanks!

natecook1000: @kylemacomber This is a more predictable version of what we've talked about for making `LazyFilter` better behaved — precalculating the start index would improve things there as well.

timvermeulen: I was thinking about this precomputation vs expensive `startIndex` vs expensive `self[startIndex]` trade-off, and I didn't know whether effectively precomputing `startIndex` was deemed "lazy enough" for lazy collections. I like this solution the most, to be clear 🙂 But at the same time, I've seen people (including myself) get confused by similar behaviour.

kylemacomber: I _think_ what I like about this approach is that it's more robust in the sense that we never end up with accidentally quadratic implementations.

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Done!

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: To me the best now is that iterations(which may be the most common operations in collections) are improved with this approach. Also it simplified the implementation which is always nice =]

LucianoPAlmeida: I've made all the adjustments so I think this is ready for review again now :))

CTMacUser: I’ve thought of caching `startIndex` for my filter-like collection wrappers before, but one part always bugged me enough to dissuade committing: what if the base collection is a reference type and it changed underneath me?!  The cached index may be invalid relative to the new base state; dereferencing may now even cause a crash.  I know 99.9% of the type the base collection will be a value type, and 99.9% of the time the base collection won’t be mutated while a chunk is active, so your solution of “so sad, too bad” is OK, but it’s still lousy from a purity perspective. The actual solution would be changing the language so general notifications are made when a reference object is mutated (so in this case, the cached `startIndex` can be recomputed), but that’s severe and I’m not sure it’s implementable.

CTMacUser:
Looking at the current code, it should be easy to set the `SubSequence` to be a wrapper over `Base.SubSequence` (instead of the default `Slice<Self>`).  Maybe I’ll try it out later.

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: I think we have a point here... also I was thinking about this and this wouldn't be a big problem if we just compute the `startIndex` every time, as imagine most cases people base would likely to be an array or an lazy collection from a array source with `RandomAccessCollection` conditional conformance where base is random access so, in most cases `startIndex` wouldn't be expensive to compute. Also, I still like this range based approach because as Kyle mentioned is more robust and iterations are improved. 

> The actual solution would be changing the language so general notifications are made when a reference object is mutated (so in this case, the cached startIndex can be recomputed), but that’s severe and I’m not sure it’s implementable.

In C# we have a concept of `ObservableCollection` where mutations(well defined though an API) can be observed from anywhere. This not guarantee a notification of a custom collection mutation through a custom API, but I just remembered the concept, which is nice and very useful in the context of .NET framework =]

kylemacomber: For a reference-type collection or a collection of reference types, if there is a mutation between when a filter-like adapter is created and when it's traversed, I agree the result could be surprising and it's worth thinking carefully about our policy.

Note that for `chunks(ofCount:)`, mutations to reference type elements in a collection shouldn't change the behavior, and mutations to a reference type collection itself, possibly could be made to work. Though depending on our policy, we may decide we don't need to be defensive here.

timvermeulen: More concretely, my only concern with this approach was that certain lazy collections (most notably `LazyFilterCollection`) will run closures and evaluate elements when an index is advanced, and so precomputing the range of the first chunk can possibly run code that the user may not expect to be run at that stage. (Counterpoint: while lazy sequences are useful for short-circuiting, the user is very likely to at least evaluate the first element, in which case precomputing the start index doesn't have much of an impact.)

Calling `.prefix(n)` on a lazy collection has the same behavior, so we might have to conclude that such things are unavoidable/acceptable 🙂 In this case I think we should consider precomputing the first chunk of `chunked(on:)` and `chunked(by:)` as well.

LucianoPAlmeida: > Calling `.prefix(n)` on a lazy collection has the same behavior, so we might have to conclude that such things are unavoidable/acceptable 🙂 In this case I think we should consider precomputing the first chunk of `chunked(on:)` and `chunked(by:)` as well.

I agree, and I think at the end of the day as @kylemacomber mentioned it all depends if we decide to be really strict and  defensive or trade that of for performance :) 
Also, I still don't think that it would be a big problem to always compute the start index if we really want to be conservative. In either way let me know what you guys think and I'll be happy to make the changes if necessary =]

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,258 @@ extension Collection {
...
+    // calculation of offsets(that is already done at `index(before:)`)
+    // we just move one position already so the index can be calculated
+    // since all remaining chunks have the same size.
+    if i.baseRange.lowerBound == base.endIndex {
+      formIndex(before: &idx)
+      distance += 1
+      // If the offset was simply one, we are done.
+      guard distance != 0 else {
+        return idx
+      }
```
timvermeulen, line 414: The `makeOffsetIndex` code below will still work if `distance` is zero, right? In that case I would suggest getting rid of these lines to make it a bit simpler.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,258 @@ extension Collection {
...
+      limitedBy: baseBound
+    ) ?? baseBound
+    
+    let baseEndIdx = base.index(
+      i.baseRange.lowerBound, offsetBy: (distance + 1) * chunkCount,
+      limitedBy: base.endIndex
+    ) ?? base.endIndex
+    
+    return Index(_baseRange: baseStartIdx..<baseEndIdx)
+  }
```
timvermeulen, line 438: This is very tricky to get right, but ideally we would only traverse over the base indices between `i` and the output index once, not twice. We even shouldn't need to traverse the slice corresponding to `i.baseRange`, because we already know how big this chunk is.

Your code is completely correct and it matches the logarithmic performance characteristics of these methods, so there's no pressure to improve on this in this PR. 🙂

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,258 @@ extension Collection {
...
+  public typealias Element = Base.SubSequence
+  
+  @usableFromInline
+  internal let base: Base
+  
+  @usableFromInline
+  internal let chunkCount: Int
+  
+  @usableFromInline
+  internal var computedStartIndex: Index
```
timvermeulen, line 274: Very minor, but possibly worth it: the only thing we need to precompute is the upper bound of the first chunk, not the lower bound as well (i.e. `base.startIndex`). Only storing the upper bound slightly reduces the size of a `ChunkedByCount` instance.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,258 @@ extension Collection {
...
+  }
+}
+
+extension ChunkedByCount.Index: Comparable {
+  @inlinable
+  public static func < (lhs: ChunkedByCount.Index,
+                        rhs: ChunkedByCount.Index) -> Bool {
+    lhs.baseRange.lowerBound < rhs.baseRange.lowerBound
+  }
+}
```
timvermeulen, line 339: `==` should also only need to compare the lower bounds.

Sources/Algorithms/Chunked.swift
CTMacUser: > The actual solution would be changing the language so general notifications are made when a reference object is mutated (so in this case, the cached `startIndex` can be recomputed), but that’s severe and I’m not sure it’s implementable.

Another solution would be limiting the `Base` to be a value type. I think this was one of my inspirations to request adding an `AnyValue` poison protocol.

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Thanks, fixed!

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: That make sense, fixed!

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,258 @@ extension Collection {
...
+    
+    let baseIdx = base.index(
+      i.baseRange.lowerBound, offsetBy: -offset,
+      limitedBy: base.startIndex
+    ) ?? base.startIndex
+    return Index(_baseRange: baseIdx..<i.baseRange.lowerBound)
+  }
+  
+  @inlinable
+  public func index(
```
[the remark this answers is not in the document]
LucianoPAlmeida: TBH the use of `formIndex(before:)` was only for trying to simplify the code and reuse the calculation of remainder. In short it was only me not wanting to manually calculate the index taking into account the remainder =] 
I just made the changes already, the code didn't look as ugly as I thought it would be 😆

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Not necessary anymore :)

Sources/Algorithms/Chunked.swift
LucianoPAlmeida: Nice catch, just fixed! Thanks :)

CTMacUser:
Tried doing turning `ChunkedByCount<T.SubSequence>` into `SubSequence`.  I forgot that a collection and its `SubSequence` must use the same `Index` type.  This is incompatible with a custom `Index` whose arity is per-instantiation, so I had to pull the `Index` type out first.

I used a Git(Hub) client that pulled directly from this PR.  I can't upload back to it, of course.  So I have no idea to express my changes as proper patches to your work.  Worse, before I tried to upload, you pushed down one of your own.  (I started an hour before you did.)  Here's a copy/paste of that whole section "Chunked.swift":

```swift
// chunks(ofCount:)

/// An index for an element (or past-the-end) of a `ChunkedByCount` collection.
public struct ChunkedByCountIndex<Base: Comparable> {
  // Using a separate type, instead of one embedded in ChunkedByCount, allows
  // ChunkedByCount<T> and ChunkedByCount<T.SubSequence> to share indices.

  @usableFromInline
  internal let baseRange: Range<Base>

  @usableFromInline
  internal init(_baseRange: Range<Base>) {
    self.baseRange = _baseRange
  }
}

extension ChunkedByCountIndex: Equatable {}

extension ChunkedByCountIndex: Comparable {
  public static func < (lhs: Self, rhs: Self) -> Bool {
    // Intervals are harder to do straight comparisons for.  But any cross-
    // overlap between non-identical ranges means the operands weren't from
    // compatible collections to begin with, a precondition violation.
    //
    // (Similar reasoning applies for not bothering to do a customized defintion
    // for the equality operator.)
    return lhs.baseRange.lowerBound < rhs.baseRange.lowerBound
  }
}

extension ChunkedByCountIndex: Hashable where Base: Hashable {}

/// A collection that presents the elements of its base collection
/// in `SubSequence` chunks of any given count.
///
/// A `ChunkedByCount` is a lazy view on the base Collection, but it does not implicitly confer
/// laziness on algorithms applied to its result.  In other words, for ordinary collections `c`:
///
/// * `c.chunks(ofCount: 3)` does not create new storage
/// * `c.chunks(ofCount: 3).map(f)` maps eagerly and returns a new array
/// * `c.lazy.chunks(ofCount: 3).map(f)` maps lazily and returns a `LazyMapCollection`
public struct ChunkedByCount<Base: Collection> {
  
  public typealias Element = Base.SubSequence
  
  @usableFromInline
  internal let base: Base
  
  @usableFromInline
  internal let chunkCount: Int
  
  @usableFromInline
  internal var startUpperBound: Base.Index

  /// Creates a view instance that presents the elements of the given collection
  /// as `SubSequence` chunks with the given count, caching the given index for
  /// where the first chunk ends.
  ///
  /// - Complexity: O(1)
  @inlinable
  internal init(_base: Base, _chunkCount: Int, _firstChunkEnd: Base.Index) {
    self.base = _base
    self.chunkCount = _chunkCount
    self.startUpperBound = _firstChunkEnd
  }
  ///  Creates a view instance that presents the elements of `base`
  ///  in `SubSequence` chunks of the given count.
  ///
  /// - Complexity: O(n)
  @inlinable
  internal init(_base: Base, _chunkCount: Int) {
    // Compute the start index upfront in order to make
    // start index a O(1) lookup.
    let baseEnd = _base.index(_base.startIndex, offsetBy: _chunkCount,
                              limitedBy: _base.endIndex) ?? _base.endIndex
    self.init(_base: _base, _chunkCount: _chunkCount, _firstChunkEnd: baseEnd)
  }
}

extension ChunkedByCount: Collection {
  public typealias Index = ChunkedByCountIndex<Base.Index>
  public typealias SubSequence = ChunkedByCount<Base.SubSequence>

  /// - Complexity: O(1)
  @inlinable
  public var startIndex: Index {
    Index(_baseRange: base.startIndex..<startUpperBound)
  }
  @inlinable
  public var endIndex: Index {
    Index(_baseRange: base.endIndex..<base.endIndex)
  }
  
  /// - Complexity: O(1)
  public subscript(i: Index) -> Element {
    precondition(i < endIndex, "Index out of range")
    return base[i.baseRange]
  }
  public subscript(bounds: Range<Index>) -> SubSequence {
    let lowerRange = bounds.lowerBound.baseRange,
        upperRange = bounds.upperBound.baseRange
    let resultBounds: Range<Base.Index>
    if lowerRange.lowerBound < upperRange.lowerBound {
      resultBounds = lowerRange.lowerBound ..< upperRange.upperBound
    } else {
      resultBounds = lowerRange.lowerBound ..< upperRange.lowerBound
      assert(resultBounds.isEmpty)
    }
    return SubSequence(_base: base[resultBounds], _chunkCount: chunkCount,
                       _firstChunkEnd: lowerRange.upperBound)
  }
  
  @inlinable
  public func index(after i: Index) -> Index {
    precondition(i < endIndex, "Advancing past end index")
    let baseIdx = base.index(
      i.baseRange.upperBound, offsetBy: chunkCount,
      limitedBy: base.endIndex
    ) ?? base.endIndex
    return Index(_baseRange: i.baseRange.upperBound..<baseIdx)
  }
}

extension ChunkedByCount:
  BidirectionalCollection, RandomAccessCollection
where Base: RandomAccessCollection {
  @inlinable
  public func index(before i: Index) -> Index {
    precondition(i > startIndex, "Advancing past start index")
    
    var offset = chunkCount
    if i.baseRange.lowerBound == base.endIndex {
      let remainder = base.count%chunkCount
      if remainder != 0 {
        offset = remainder
      }
    }
    
    let baseIdx = base.index(
      i.baseRange.lowerBound, offsetBy: -offset,
      limitedBy: base.startIndex
    ) ?? base.startIndex
    return Index(_baseRange: baseIdx..<i.baseRange.lowerBound)
  }
}

extension ChunkedByCount {
  @inlinable
  public func distance(from start: Index, to end: Index) -> Int {
    let distance =
      base.distance(from: start.baseRange.lowerBound,
                    to: end.baseRange.lowerBound)
    let (quotient, remainder) =
      distance.quotientAndRemainder(dividingBy: chunkCount)
    return quotient + remainder.signum()
  }

  @inlinable
  public var count: Int {
    let (quotient, remainder) =
      base.count.quotientAndRemainder(dividingBy: chunkCount)
    return quotient + remainder.signum()
  }
  
  @inlinable
  public func index(
    _ i: Index, offsetBy offset: Int, limitedBy limit: Index
  ) -> Index? {
    guard offset != 0 else { return i }
    guard limit != i else { return nil }
    
    if offset > 0 {
      guard limit < i || distance(from: i, to: limit) >= offset else {
        return nil
      }
      return offsetForward(i, offsetBy: offset)
    } else {
      guard limit > i || distance(from: i, to: limit) <= offset else {
        return nil
      }
      return offsetBackward(i, offsetBy: offset)
    }
  }

  @inlinable
  public func index(_ i: Index, offsetBy distance: Int) -> Index {
    guard distance != 0 else { return i }
    
    return distance > 0
        ? offsetForward(i, offsetBy: distance)
        : offsetBackward(i, offsetBy: distance)
  }
  
  @usableFromInline
  internal func offsetForward(_ i: Index, offsetBy distance: Int) -> Index {
    assert(distance > 0)
    return makeOffsetIndex(
      from: i, baseBound: base.endIndex, baseDistance: distance * chunkCount
    )
  }
  
  @usableFromInline
  internal func offsetBackward(_ i: Index, offsetBy distance: Int) -> Index {
    assert(distance < 0)
    if i.baseRange.lowerBound == base.endIndex {
      let remainder = base.count%chunkCount
      // We have to take it into account when calculating offsets.
      if remainder != 0 {
        // Distance "minus" one(at this point distance is negative) because we
        // need to adjust for the last position that have a variadic(remainder)
        // number of elements.
        let baseDistance = ((distance + 1) * chunkCount) - remainder
        return makeOffsetIndex(
          from: i, baseBound: base.startIndex, baseDistance: baseDistance
        )
      }
    }
    return makeOffsetIndex(
      from: i, baseBound: base.startIndex, baseDistance: distance * chunkCount
    )
  }
  
  // Helper to compute index(offsetBy:) index.
  @inline(__always)
  private func makeOffsetIndex(
    from i: Index, baseBound: Base.Index, baseDistance: Int
  ) -> Index {
    let baseStartIdx = base.index(
      i.baseRange.lowerBound, offsetBy: baseDistance,
      limitedBy: baseBound
    ) ?? baseBound
    
    let baseEndIdx = base.index(
      baseStartIdx, offsetBy: chunkCount, limitedBy: base.endIndex
    ) ?? base.endIndex
    
    return Index(_baseRange: baseStartIdx..<baseEndIdx)
  }
}

extension Collection {
  /// Returns a `ChunkedCollection<Self>` view presenting the elements
  /// in chunks with count of the given count parameter.
  ///
  /// - Parameter size: The size of the chunks. If the count parameter
  ///   is evenly divided by the count of the base `Collection` all the
  ///   chunks will have the count equals to size.
  ///   Otherwise, the last chunk will contain the remaining elements.
  ///
  ///     let c = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
  ///     print(c.chunks(ofCount: 5).map(Array.init))
  ///     // [[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]]
  ///
  ///     print(c.chunks(ofCount: 3).map(Array.init))
  ///     // [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10]]
  ///
  /// - Complexity: O(1)
  @inlinable
  public func chunks(ofCount count: Int) -> ChunkedByCount<Self> {
    precondition(count > 0, "Cannot chunk with count <= 0!")
    return ChunkedByCount(_base: self, _chunkCount: count)
  }
}

// Conditional conformances.
extension ChunkedByCount: Equatable where Base: Equatable {}

// Since we have another stored property of type `Index` on the
// collection, synthesis of `Hashble` conformace would require
// a `Base.Index: Hashable` constraint, so we implement the hasher
// only in terms of `base`. Since the computed index is based on it,
// it should not make a difference here.
extension ChunkedByCount: Hashable where Base: Hashable {
  public func hash(into hasher: inout Hasher) {
    hasher.combine(base)
  }
}

// Lazy conditional conformance.
extension ChunkedByCount: LazySequenceProtocol
  where Base: LazySequenceProtocol {}
extension ChunkedByCount: LazyCollectionProtocol
  where Base: LazyCollectionProtocol {}
```

I didn't touch the test file, so you have to add something that executes the subsequence subscript method.

timvermeulen:
@CTMacUser I like the idea of this — `Slice<ChunkedBySize<Base>>` effectively stores 5 instances of `B.Index`, whereas `ChunkedBySize<Base.SubSequence>` only store 1 (plus however many `Base.SubSequence` stores, which is 2 in the case of `Slice<Base>`). I'm just curious if this is also the upside you had in mind or that this has other benefits as well.

Either way, this might be more suitable as a follow-up PR since this PR seems to be good to go, and it seems like the `LazyChunked` collection could benefit from this as well?

> I forgot that a collection and its `SubSequence` must use the same `Index` type.

Thanks for pointing this out, I'm making a note that `validateIndexTraversals` should probably test this somehow 🙂

natecook1000:
> Either way, this might be more suitable as a follow-up PR since this PR seems to be good to go, and it seems like the LazyChunked collection could benefit from this as well?

Agreed, this is a cool change, but let's get this integrated first.

natecook1000 (approved):
Looks good!

Sources/Algorithms/Chunked.swift
timvermeulen: This looks like a good improvement, though this is still doing some unnecessary work (just less than before 😄). Offsetting an index forwards would ideally offset the upper bound of the index's base range, and not the lower bound. e.g. if you offset `Index(a..<b)` by `n`, then the lower bound of the output range is `(n - 1) * chunkSize` away from `b`, which is less work to compute (for non-random-access collections) than the index `n * chunkSize` away from `a`. And similarly, when offsetting an index backwards you would ideally first compute the upper bound of the output range and then the lower bound, rather than the other way around.

None of this matters much for random-access collections, but preferably on non-random-access collections these methods don't do more work than the naive solution of repeatedly calling `index(after:)` / `index(before:)`.

There doesn't seem to be a very easy way to fix this because you use `makeOffsetIndex` for both forward and backward offsets. Just to reiterate, this doesn't need to be perfect on the first go — there's a reason I still haven't finished my index offset logic for the `Product2` type 🙂 This stuff can be quite hairy.

Sources/Algorithms/Chunked.swift
```diff
@@ -246,3 +246,261 @@ extension Collection {
...
+      guard limit < i || distance(from: i, to: limit) >= offset else {
+        return nil
+      }
+      return offsetForward(i, offsetBy: offset)
+    } else {
+      guard limit > i || distance(from: i, to: limit) <= offset else {
+        return nil
+      }
+      return offsetBackward(i, offsetBy: offset)
+    }
```
timvermeulen, line 404: These `distance(from:to:)` calls aren't desirable here because they're potentially very expensive on non-random-access collections (they might end up having to iterate the entire collection). I [solved](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/SlidingWindows.swift#L90-L127) this for `SlidingWindows` by passing the limit to the respective `offsetForward` and `offsetBackward` methods.

LucianoPAlmeida: Just fixed, thanks

LucianoPAlmeida:
@natecook1000 Sorry,  I've sent some updates can you give another look? :)

CTMacUser:
> Either way, this might be more suitable as a follow-up PR since this PR seems to be good to go, and it seems like the `LazyChunked` collection could benefit from this as well?

Since your post, this PR has been merged and withdrawn.  So I have to figure out how to resubmit the changes.

> > I forgot that a collection and its `SubSequence` must use the same `Index` type.
> 
> Thanks for pointing this out, I'm making a note that `validateIndexTraversals` should probably test this somehow 🙂

It's a protocol requirement; the compiler will block attempts for `SubSequence` to use a different `Index` type by reporting non-conformance.  So you wouldn't need an extra test.

PR #55, opened 2021-01-05
Add `lastIndex` variants returning `Range<Index>?`

### LINKED ISSUE #23: Add `indexBeforeFirst(where:)` and `indexAfterLast(where:)`

benrimmington:
### Description

Closes #23

### Detailed Design

```swift
  public func lastIndexAsRange(
    where predicate: (Element) throws -> Bool
  ) rethrows -> Range<Index>?
}

  public func lastIndexAsRange(
    of element: Element
  ) -> Range<Index>?
}
```

### Documentation Plan

- README.md (added link to guide).
- Guides/LastIndex.md (added example code).
- Sources/Algorithms/LastIndex.swift (added documentation comments).

### Test Plan

- Tests/SwiftAlgorithmsTests/LastIndexTests.swift

### Source Impact

None (new API only).

### Checklist

### DIFF
--- a/Guides/LastIndex.md
+++ b/Guides/LastIndex.md
@@ -1,0 +1,25 @@
+# LastIndex
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/LastIndex.swift) |
+  [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/LastIndexTests.swift)]
+
+The `lastIndexAsRange(where:)` and `lastIndexAsRange(of:)` methods return an optional range,
+containing the last index of a matching element in the bidirectional collection.
+
+The range's upper bound can then be used to form another half-open range.
+
+```swift
+let hexDigits: [Character] = Array("0123456789ABCDEF")
+
+hexDigits.lastIndexAsRange(where: \.isLetter)  //-> 0xF..<0x10
+hexDigits.lastIndexAsRange(where: \.isNumber)  //-> 0x9..<0xA
+hexDigits.lastIndexAsRange(where: \.isSymbol)  //-> nil
+hexDigits.lastIndexAsRange(of: "F")            //-> 0xF..<0x10
+hexDigits.lastIndexAsRange(of: "9")            //-> 0x9..<0xA
+hexDigits.lastIndexAsRange(of: "$")            //-> nil
+
+if let range = hexDigits.lastIndexAsRange(where: \.isNumber) {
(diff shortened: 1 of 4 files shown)


xwu:
How about just `lastIndexRange`?

benrimmington:
> How about just `lastIndexRange`?

@xwu I rejected `lastIndexRange` and `lastIndices`, because they could be misinterpreted as returning a range containing (potentially) more than one index.

xwu:
> because they could be misinterpreted as returning a range containing (potentially) more than one index.

Wait...it _doesn't_ do that? Well, now you have empiric evidence that this API has been misinterpreted with its current naming anyway.

It seem kind of a niche operation: What's the use case? Can we come up with something that can be re-used more widely which would serve the same use case?

benrimmington:
> It seem kind of a niche operation: What's the use case?

* <https://github.com/apple/swift-algorithms/issues/23#issue-721433471> is the original use case.

* <https://github.com/apple/swift-algorithms/issues/23#issuecomment-754913814> is the negative feedback for this PR.

xwu:
Got it: I think I would agree with @natecook1000 that, if in the predominant use cases you tend to want just one index or the other, then the API would be less cumbersome returning the one thing you actually want, at the expense of requiring another operation in the rare case you want both.

PR #56, opened 2021-01-11
All permutations

mdznr:
### Description
This addition adds the ability to easily get permutations of a range of sizes, instead of the previous behavior of a single, fixed size, `k`, like https://github.com/apple/swift-algorithms/pull/51 did for `Combinations`.

### Detailed Design
The `permutations(ofCount:)` function that takes a `RangeExpression` is added in an extension of `Collection`.

```swift
extension Collection {
  public func permutations<R: RangeExpression>(
    ofCount kRange: R
  ) -> Permutations<Self> where R.Bound == Int {
    return Permutations(self, kRange: kRange)
  }
}
```

If the upper bound of the range exceeds the length of the collection, no such permutations of those sizes will be returned, effectively clamping the range to `0...n` where `n` is the length of the collection.

### Naming
The name `permutations(ofCount:)` is the same as the original, but reflects the ability to provide multiple counts. This is [the same pattern used for `Combinations`](https://github.com/apple/swift-algorithms/pull/51).

### Documentation Plan
The documentation in `Guides/Permutations.md` has been updated to reflect the new function. Inline code documentation also exists for each of the functions.

### Test Plan
- Adds tests for the outputs of a range of sizes
- Adds tests for ranges extending above the size of the collection

### Source Impact
The functionality is purely additive. The behavior of the existing function is unchanged.

### Checklist

### DIFF
--- a/Sources/Algorithms/Permutations.swift
+++ b/Sources/Algorithms/Permutations.swift
@@ -149,11 +149,8 @@ extension Permutations: Sequence {
       /// Advances `kRange` by incrementing its `lowerBound` until the range is
       /// empty, when iteration is finished.
       func advanceKRange() {
-        if kRange.lowerBound < kRange.upperBound {
-          let advancedLowerBound = kRange.lowerBound + 1
-          kRange = advancedLowerBound ..< kRange.upperBound
-          indexes = Array(base.indices)
-        }
+        kRange.removeFirst()
+        indexes = Array(base.indices)
       }
       
       let countToChoose = self.kRange.lowerBound
--- a/Sources/Algorithms/Permutations.swift
+++ b/Sources/Algorithms/Permutations.swift
@@ -301,7 +301,9 @@ extension Collection {
   ///   If `k` is `nil`, the resulting sequence represents permutations of this
   ///   entire collection.
   ///
-  /// - Complexity: O(1)
+  /// - Complexity: O(1) for random-access base collections. O(*n*) where *n*
+  /// is the number of elements in the base collection, since `Permutations`
+  /// accesses the `count` of the base collection.
   @inlinable
   public func permutations(ofCount k: Int? = nil) -> Permutations<Self> {
     precondition(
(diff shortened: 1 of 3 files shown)


kylemacomber:
@mdznr Oh I see this is still a draft, let us know when you're ready for us to review it!

mdznr:
> @mdznr Oh I see this is still a draft, let us know when you're ready for us to review it!

I unfortunately haven’t had a lot of time to work on this. The change is very nearly done. It works. The only things left for me was to ensure it was as well-written as it could be, the implementation consistent with `Combinations` as much as possible, and the set of unit tests are thorough. I was hoping to really dive into those and ensure it was up to standards before asking others to review.

Sources/Algorithms/Permutations.swift
```diff
@@ -82,72 +113,95 @@ extension Permutations: Sequence {
...
+      guard !isFinished else { return nil }
       
+      /// Advances `kRange` by incrementing its `lowerBound` until the range is
+      /// empty, when iteration is finished.
+      func advanceKRange() {
+        if kRange.lowerBound < kRange.upperBound {
+          let advancedLowerBound = kRange.lowerBound + 1
+          kRange = advancedLowerBound ..< kRange.upperBound
+          indexes = Array(base.indices)
+        }
```
timvermeulen, line 156:
```suggestion
        kRange.removeFirst()
        indexes = Array(base.indices)
```

You've already checked that `!isFinished` (i.e. `!kRange.isEmpty`), so no need to check that here.

timvermeulen (changes requested):
Thanks a lot for your contribution, @mdznr! This looks very solid, I just have one remark. Feel free to make any other changes you still want to make, but I'm down to get this merged in its current state (modulo that remark).

Sources/Algorithms/Permutations.swift
mdznr: Good catch. Thank you!

timvermeulen (approved):

Issue #62, opened 2021-01-16
Add a suffix(while:) method


natecook1000:
The standard library includes `prefix` and `suffix` methods that take a number of elements or an index, but only a `prefix` method that takes a predicate. We should add a `suffix(while:)` method to match `prefix(while:)` for bidirectional collections.

michiboo:
Hi I would like to take this issue!  where should I add suffix(while:) methods?

kylemacomber:
Please open a PR against this package. It probably makes sense to add a new file, say `Sources/Algorithms/Suffix.swift`, and add your implementation there.

PR #65, opened 2021-01-20
add suffix:while method 

michiboo:
Thanks for contributing to Swift Algorithms!

    
    - [x] I've completed this task
    - [ ] This task isn't completed

Add a suffix(while:) method #62
### Checklist
- [ ] I've read the [Contribution Guidelines](../CONTRIBUTING.md)


Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// Suffix(while:)
+//===----------------------------------------------------------------------===//
+
+extension BidirectionalCollection {
+  @inlinable
+  public __consuming func Suffix(while predicate: (Element) throws -> Bool
```
mdznr, line 18:
```suggestion
  public __consuming func suffix(while predicate: (Element) throws -> Bool
```

mdznr, line 18:
```suggestion
  public __consuming func suffix(
    while predicate: (Element) throws -> Bool
```

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
...
+  ) rethrows -> [Element] {
```
mdznr, line 19: Can this return `SubSequence` instead?

`prefix(while:)` on `Collection` returns a `SubSequence`.<sup>[1](https://github.com/apple/swift/blob/main/stdlib/public/core/Collection.swift#L1328)</sup>

----

1. https://github.com/apple/swift/blob/main/stdlib/public/core/Collection.swift#L1328

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// Suffix(while:)
```
mdznr, line 13:
```suggestion
// suffix(while:)
```

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
...
+//===----------------------------------------------------------------------===//
+
+extension BidirectionalCollection {
+  @inlinable
```
mdznr, line 17: Documentation should be added. Mirroring the documentation for `prefix(while:)` would be a good start.

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
...
+   var result = ContiguousArray<Element>()
+   self.reverse()
+    for element in self {
+      guard try predicate(element) else {
+        break
+      }
+      result.append(element)
+    }
+    result.reverse()
+    return Array(result)
```
mdznr, line 29: This double-reverse operation and converting to `Array` shouldn’t be necessary. The last index of the collection can be retrieved with `endIndex`, then walked backwards using `index(before:)`.

This would be similar to the implementation of `prefix(while:)`<sup>[1](https://github.com/apple/swift/blob/main/stdlib/public/core/Collection.swift#L1330)</sup>, but a little bit trickier. Since `endIndex` is beyond the subscript-able range of a collection, the index has to be moved backwards _before_ using it to subscript the collection, and ensuring it doesn’t go before the first element (start index). In SuffixTests.swift, I made some comments about test cases that will be important for ensuring that these cases are accounted for.

----

1. https://github.com/apple/swift/blob/main/stdlib/public/core/Collection.swift#L1330

Tests/SwiftAlgorithmsTests/SuffixTests.swift
```diff
@@ -0,0 +1,20 @@
...
+
+import XCTest
+import Algorithms
+
+final class SuffixTests: XCTestCase {
+  
+  func testSuffix() {
+    let a = 0...10
+    XCTAssertEqualSequences(a.suffix(while: { $0 > 5 }), (6...10))
+}
```
mdznr, line 20:
```suggestion
  }
}
```

Tests/SwiftAlgorithmsTests/SuffixTests.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import Algorithms
+
+final class SuffixTests: XCTestCase {
+
```
mdznr, line 16:
```suggestion
```

Tests/SwiftAlgorithmsTests/SuffixTests.swift
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import Algorithms
+
+final class SuffixTests: XCTestCase {
```
mdznr, line 15: Tests should be added to handle:
- Empty input
- Empty output (predicate doesn’t match the first element)
- Predicate matches just the first element
- Predicate matches all elements

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
```
mdznr, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the Swift project authors
```

Tests/SwiftAlgorithmsTests/SuffixTests.swift
```diff
@@ -0,0 +1,20 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
```
mdznr, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the Swift project authors
```

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,31 @@
...
+//===----------------------------------------------------------------------===//
+// Suffix(while:)
+//===----------------------------------------------------------------------===//
+
+extension BidirectionalCollection {
+  @inlinable
+  public __consuming func Suffix(while predicate: (Element) throws -> Bool
+  ) rethrows -> [Element] {
+   var result = ContiguousArray<Element>()
+   self.reverse()
```
mdznr, line 21: Consistent indentation

```suggestion
    var result = ContiguousArray<Element>()
    self.reverse()
```

mdznr, line 21: I don’t believe `reverse()` can be used here (won’t compile), because it is a `mutating` function that reverses the elements of the collection in-place. This function, `suffix(while:)`, isn’t mutating (and should remain that way), so it can’t use `self.reverse()` in its implementation.

It seems that you might have meant to use the non-mutating version of that function, `reversed()`, which returns a reversed collection:

```swift
    for element in self.reversed() {
```

However, this comment likely won’t be applicable when changing the function to return `SubSequence` instead.

mdznr (commented):
Excited to see this function added, bringing more parity between the suffix and prefix family of functions!

If changes are still being worked on, I’d recommend making use of a new feature on GitHub: Drafts. By marking this PR as a draft, it helps to convey the author’s intended state of the changes.<sup>[[1]](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request)</sup>

I left a few small comments inline, but want to emphasize the [primary high-level piece of feedback](https://github.com/apple/swift-algorithms/pull/65#discussion_r561041229) here, which may make several of the comments inapplicable. I think `suffix(while:)` should return `SubSequence` instead of `[Element]` for consistency and performance reasons.

----

1. https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request

CTMacUser:
What about something like (neither compiled nor tested):

```swift
  public func suffix(
  ) rethrows -> SubSequence {
    let start = startIndex
    var result = endIndex
    while result > start {
      let previous = index(before: result)
      guard try predicate(self[previous]) else { break }

      result = previous
    }
    return self[result...]
  }
}
```

?

mdznr:
@CTMacUser I think that’s pretty much exactly what I imagined the implementation would look like.

Although, I think we might want to change

```swift
while result > start {
```

to

```swift
while result != start {
```

A lot of similar algorithms in the standard library use `!=` instead of `>` or `<` for comparing indexes. I’ll have to do more research as to why that is. My initial guess is perhaps because implementations of custom indexes have slightly faster implementations of `!=` than `<`. But either way, it would be good to have consistency here.

Sources/Algorithms/Suffix.swift
```diff
@@ -14,18 +14,17 @@
...
+      guard try predicate(self[previous]) else { break }
+
+      result = previous
     }
-    result.reverse()
-    return Array(result)
+    return self[result...]
   }
-}
+}
```
xwu, line 30:
```suggestion
}

```

kylemacomber:
> A lot of similar algorithms in the standard library use != instead of > or < for comparing indexes. I’ll have to do more research as to why that is. My initial guess is perhaps because implementations of custom indexes have slightly faster implementations of != than <. But either way, it would be good to have consistency here.

I agree that we should use `!=` to be consistent with the stdlib.

One of the reasons stdlib uses `!=` is historical. Originally `Collection.Index` in the stdlib was only `Equatable`, not `Comparable`. From [SE-0065](https://github.com/apple/swift-evolution/blob/main/proposals/0065-collections-move-indices.md) which added the `Comparable` requirement:

> It's worth noting that this requirement isn't strictly necessary. Without it, though, indices would have no requirements beyond `Equatable`, and creation of a `Range<T>` would have to be allowed for any `T` conforming to `Equatable`. As a consequence, most interesting range operations, such as containment checks, would be unavailable unless `T` were also `Comparable`, and we'd be unable to provide bounds-checking in the general case.
>
> That said, the requirement has real benefits. For example, it allows us to support distance measurement between arbitrary indices, even in collections without random access traversal. In the old model, `x.distance(to: y)` for these collections had the undetectable precondition that `x` precede `y`, with unpredictable consequences for violation in the general case.

Sources/Algorithms/Suffix.swift
```diff
@@ -14,6 +14,15 @@
 //===----------------------------------------------------------------------===//
 
 extension BidirectionalCollection {
+  /// Returns a subsequence containing the elements from the end until `predicate`
+  /// returns `false` and skipping the remaining elements.
```
xwu, line 18:
```suggestion
  /// Returns a subsequence containing the elements from the end until
  /// `predicate` returns `false` and skipping the remaining elements.
```

Sources/Algorithms/Suffix.swift
```diff
@@ -0,0 +1,39 @@
...
+  /// - Complexity: O(*n*), where *n* is the length of the collection.
+  public func suffix(
+    while predicate: (Element) throws -> Bool
+  ) rethrows -> SubSequence {
+    let start = startIndex
+    var result = endIndex
+    while result != start {
+      let previous = index(before: result)
+      guard try predicate(self[previous]) else { break }
+
```
xwu, line 34:
```suggestion
```

natecook1000 (changes requested):
Thanks for your contribution, @michiboo! Could you update your tests to compile and pass? It looks like this is otherwise ready to go.

natecook1000:
This is all set now — thanks for this implementation, @michiboo!

PR #66, opened 2021-01-20
Index offsetting for Product2

timvermeulen:
This PR adds implementations of `index(_:offsetBy:)` and `index(_:offsetBy:limitedBy:)` to the `Product2` collection. There are ways to further reduce the amount of indices traversed, but this should perform well enough.

LIke `Chain2`, `Product2`'s index offsetting behavior will now unfortunately perform slightly worse for non-random-access collections in specific scenarios, but there's not much we can do about that.

### Checklist

### DIFF
--- a/Sources/Algorithms/Product.swift
+++ b/Sources/Algorithms/Product.swift
@@ -124,11 +124,11 @@ extension Product2: Collection where Base1: Collection {
     (base1[position.i1], base2[position.i2])
   }
   
-  /// Forms an index from a pair of base indices, converting
+  /// Forms an index from a pair of base indices, normalizing
   /// `(i, base2.endIndex)` to `(base1.index(after: i), base2.startIndex)` if
   /// necessary.
   @usableFromInline
-  internal func convertIndex(_ i1: Base1.Index, _ i2: Base2.Index) -> Index {
+  internal func normalizeIndex(_ i1: Base1.Index, _ i2: Base2.Index) -> Index {
     i2 == base2.endIndex
       ? Index(i1: base1.index(after: i1), i2: base2.startIndex)
       : Index(i1: i1, i2: i2)
@@ -137,7 +137,7 @@ extension Product2: Collection where Base1: Collection {
   @inlinable
   public func index(after i: Index) -> Index {
     precondition(i.i1 != base1.endIndex, "Can't advance past endIndex")
-    return convertIndex(i.i1, base2.index(after: i.i2))
+    return normalizeIndex(i.i1, base2.index(after: i.i2))
   }
   
   @inlinable
@@ -259,7 +259,7 @@ extension Product2: Collection where Base1: Collection {
       // i.i1 > [x x x|x x x x x x|x x x]
       //        [     |> > > > > >|     ]   (`distance`)
       
-      return convertIndex(i.i1, i2)
+      return normalizeIndex(i.i1, i2)
     }
     
     let suffixCount = base2[i.i2...].count
--- a/Sources/Algorithms/Product.swift
+++ b/Sources/Algorithms/Product.swift
@@ -9,7 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// A sequence that represents the product of two sequence's elements.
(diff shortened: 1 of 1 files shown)


Sources/Algorithms/Product.swift
```diff
@@ -119,17 +119,27 @@ extension Product2: Collection where Base1: Collection {
...
+  @inlinable
+  public subscript(position: Index) -> (Base1.Element, Base2.Element) {
+    (base1[position.i1], base2[position.i2])
+  }
+  
+  /// Forms an index from a pair of base indices, converting
+  /// `(i, base2.endIndex)` to `(base1.index(after: i), base2.startIndex)` if
+  /// necessary.
+  @usableFromInline
+  internal func convertIndex(_ i1: Base1.Index, _ i2: Base2.Index) -> Index {
```
kylemacomber, line 131: is this a "conversion" or is it more of a "normalization" or "canonicalization"?

Sources/Algorithms/Product.swift
```diff
@@ -183,10 +193,219 @@ extension Product2: Collection where Base1: Collection {
...
+  
+  public func index(
+    _ i: Index,
+    offsetBy distance: Int,
+    limitedBy limit: Index
+  ) -> Index? {
+    if distance >= 0 {
+      return limit >= i
+        ? offsetForward(i, by: distance, limitedBy: limit)
+        : offsetForward(i, by: distance)
```
kylemacomber, line 213: I found these ternaries confusing until I went and read the documentation for `index(_:offsetBy:limitedBy:)`, in particular:

```
  ///   - limit: A valid index of the collection to use as a limit. If
  ///     `distance > 0`, a limit that is less than `i` has no effect.
  ///     Likewise, if `distance < 0`, a limit that is greater than `i` has no
  ///     effect.
```

I don't know if we should add a comment inline to this effect, or if it's reasonable to assume that if you're reading an implementation of this method you should be familiar with the expected behavior.

kylemacomber (commented):
I got part way through this patch, but wasn't able to finish. I'll pick back up reviewing `offsetForward(_:by:limitedBy:)` and `offsetBackward(_:by:limitedBy:)` tomorrow.

Sources/Algorithms/Product.swift
```diff
@@ -183,10 +193,219 @@ extension Product2: Collection where Base1: Collection {
...
+    guard let i1 = base1.index(nextI1, offsetBy: base1Distance, limitedBy: limit.i1)
+      else { return nil }
+    
+    // The distance from `base2.startIndex` to the target.
+    let base2Distance = remaining % base2Count
+    
+    let base2Limit = limit.i1 == i1 ? limit.i2 : base2.endIndex
+    return base2.index(base2.startIndex, offsetBy: base2Distance, limitedBy: base2Limit)
+      .map { i2 in Index(i1: i1, i2: i2) }
+  }
```
kylemacomber, line 311: I think my main question with this patch is how much do these special cases in `offsetForward` and `offsetBackward` buy us? It's a lot to hold in your head.

Can we get away with just the main "loop"?

```swift
internal func offsetForward(
  _ i: Index, by distance: Int, limitedBy limit: Index
) -> Index? {
  assert(distance >= 0)
  assert(limit >= i)

  let suffixCount = base2[i.i2...].count
  let remaining = distance - suffixCount
  let nextI1 = base1.index(after: i.i1)
  let prefixCount = base2[..<i.i2].count
  let base2Count = prefixCount + suffixCount
  let base1Distance = remaining / base2Count

  guard let i1 = base1.index(nextI1, offsetBy: base1Distance, limitedBy: limit.i1)
    else { return nil }

  // The distance from `base2.startIndex` to the target.
  let base2Distance = remaining % base2Count

  let base2Limit = limit.i1 == i1 ? limit.i2 : base2.endIndex
  return base2.index(base2.startIndex, offsetBy: base2Distance, limitedBy: base2Limit)
    .map { i2 in Index(i1: i1, i2: i2) }
}
```

kylemacomber (approved):

PR #68, opened 2021-01-21
Precompute LazyChunked's startIndex

timvermeulen:
This PR adds a `firstUpperBound` stored property to the `LazyChunked` type to avoid computing the first chunk twice during iteration, similar to what the `ChunkedByCount` and `SlidingWindows` types do.

### Checklist


natecook1000:
We should note that the complexity of calling `chunked` is O(n) with this change (instead of O(1) like we currently have for lazy methods).

9396985334 (approved):

Issue #69, opened 2021-01-25
randomSample crashes randomly


sidepelican:
`randomSample` may have crash in case of `0` was out from `Double.random(in: 0..<1)`.
`log(0)` is invalid because the domain of `log(x)` is `0 < x`.
It seems that `Double.log(0.0)` produces `-inf`.

```
(lldb) po Double.log(0.0)
-inf
```

`-inf` cannot cast to `Int` so the next crash happens.

![](https://user-images.githubusercontent.com/19257572/105685389-5b08a580-5f39-11eb-8f4c-d20de3b21dd1.png)

**Swift Algorithms version:** `0.0.2`
**Swift version:** Paste the output of `swift --version` here
```
Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
Target: x86_64-apple-darwin19.6.0
```

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-algorithms/issues)

### Steps to Reproduce

I can reproduce this behavior from the code below.

```swift
var generator = SplitMix64(seed: 0x61c8864680b583eb) // this generator outs `0` first.
_ = c.randomSample(count: k, using: &generator) // crash!
```

### Expected behavior

Does not crash.

### Actual behavior

pasted screeenshot above.

PR #70, opened 2021-01-25
Fix RandomSample crash when 0 was out from random.

sidepelican:
Fix https://github.com/apple/swift-algorithms/issues/69 .

I changed the random ranges from `[0, 1)` to `(0, 1]` .
`w` must not be `0` so fixed `nextW` too. (in `log(1 - w)`, `log(1)` is `0` so zero division happens.)

### Checklist

### DIFF
--- a/Sources/Algorithms/RandomSample.swift
+++ b/Sources/Algorithms/RandomSample.swift
@@ -87,14 +87,15 @@ extension Collection {
 internal func nextW<G: RandomNumberGenerator>(
   k: Int, using rng: inout G
 ) -> Double {
-  Double.root(1 - .random(in: 0..<1, using: &rng), k)
+  Double.root(.random(in: 0..<1, using: &rng), k)
 }
 
 @usableFromInline
 internal func nextOffset<G: RandomNumberGenerator>(
   w: Double, using rng: inout G
 ) -> Int {
-  Int(Double.log(1 - .random(in: 0..<1, using: &rng)) / .log(1 - w))
+  let offset = Double.log(.random(in: 0..<1, using: &rng)) / .log(onePlus: -w)
+  return offset < Double(Int.max) ? Int(offset) : Int.max
 }
 
 extension Collection {
@@ -201,7 +202,7 @@ extension Sequence {
       w *= nextW(k: k, using: &rng)
       
       // Find the offset of the next element to swap into the reservoir.
-      var offset = nextOffset(w: w, using: &rng) + 1
+      var offset = nextOffset(w: w, using: &rng) &+ 1
       
       // Skip over `offset - 1` elements to find the selected element.
       while offset > 1, let _ = iterator.next() {
--- a/Sources/Algorithms/RandomSample.swift
+++ b/Sources/Algorithms/RandomSample.swift
@@ -87,14 +87,14 @@ extension Collection {
 internal func nextW<G: RandomNumberGenerator>(
   k: Int, using rng: inout G
 ) -> Double {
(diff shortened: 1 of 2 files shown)


timvermeulen:
Thanks for your contribution, @sidepelican! This looks great.

`1 - .random(in: 0..<1, using: &rng)` is a good way of excluding `0` from the result. It does now include the upper bound, `1`, which the original algorithm does not, but as far as I can see this does not affect the algorithm in any meaningful way.

Your seeded number generator correctly produces `0` as its first random number, and this tests the change you made in `nextW` (which is where `rng` is first used). However, note that it does not test the change in `nextOffset` because at that point `rng` no longer  generates `0`. Can you think of a way to test both changes, perhaps using a custom `RandomNumberGenerator` type?

sidepelican:
> Can you think of a way to test both changes, perhaps using a custom RandomNumberGenerator type?

I can make the `RandomNumberGenerator` type its anytime returns `0`, but it doesn't work correctly becase `Int.random(in: 0..<N, using: &g)` (N > 2) falls into infinit loop.

https://github.com/apple/swift-algorithms/blob/cda6fddada47e3146fdfc06504e1d8b2d5ff20dd/Sources/Algorithms/RandomSample.swift#L140
The test stops at this line when using such a generator.

The fundamental loop is here.
https://github.com/apple/swift/blob/a73a8087968f9111149073107c5242d83635107a/stdlib/public/core/Random.swift#L104

----

Based on the above, I changed the test to use internal function directly. 5acda56
How about this? @timvermeulen

timvermeulen (approved):
> I can make the `RandomNumberGenerator` type its anytime returns `0`, but it doesn't work correctly becase `Int.random(in: 0..<N, using: &g)` (N > 2) falls into infinit loop.

Indeed, what I had in mind was either a generator that starts by generating at least two `0`s like you mentioned, or one that generates a `0` with a certain (high) probability.

> Based on the above, I changed the test to use internal function directly.

This I like even more! Now we're not making any assumptions about the exact order in which these functions are called.

stephentyrone:
I think this means that we can't test in release any longer, right? @natecook1000, how do you feel about that?

stephentyrone:
We chatted, and we're fine with this. Approved.

stephentyrone (approved):

stephentyrone:
Upon more detailed review, this fixes the most obvious problem, but is not actually a complete solution. `log(1-w)` can underflow to zero without `w` being exactly zero; we simply need to have a sequence of `w` values whose product is less than `.ulpOfOne/2`. This should be replaced with `log(onePlus: -w)` to resolve the problem.

timvermeulen:
Nice catch, and TIL about `log(onePlus:)`!

> numerically this only matters for very large collections--more than 2**53 elements

We can trivially construct a range larger than that, so we have real reasons to fix this.

sidepelican:
TIL about `log(onePlus:)` too!

I tried https://github.com/apple/swift-algorithms/pull/70#issuecomment-774701048 , its looks fine but, I found crash when returning `Int.max` in `nextOffset` because arithmetic overflow happens at here.

https://github.com/apple/swift-algorithms/blob/cda6fddada47e3146fdfc06504e1d8b2d5ff20dd/Sources/Algorithms/RandomSample.swift#L204

I fixed this by just adding `&` but I'm not smart about the algorithm there so is this right?

stephentyrone:
Can't we eliminate the +1 there and use `while offset > 0` in the loop following?

PR #73, opened 2021-01-31
Add Sequence.uniqued(on:) which uses Key Paths

otaviocc:
### Description

This PR adds a new method to Sequence, `uniqued(on:)`. The method uses a key path to determine the uniqueness of elements in the array.

### Detailed Design

The change adds the following API:

```swift
public func uniqued<Subject: Hashable>(
    on keyPath: KeyPath<Element, Subject>
) -> [Element]
```

### Documentation Plan

Documentation was added to the code and to the guide (`Guides/Unique.md`).

### Test Plan

The method was tested using the same example used by 

```swift
public func uniqued<Subject: Hashable>(
    on projection: (Element) throws -> Subject
) rethrows -> [Element]
```

```swift
let a = ["Albemarle", "Abeforth", "Astrology", "Brandywine", "Beatrice", "Axiom"]
let b = a.uniqued(on: { $0.first })
XCTAssertEqual(["Albemarle", "Brandywine"], b)

let c = a.uniqued(on: \.first)
XCTAssertEqual(["Albemarle", "Brandywine"], d)
```

### Source Impact

Since it's a new method, there's no impact on existing users.

### Checklist

### DIFF
--- a/Guides/Unique.md
+++ b/Guides/Unique.md
@@ -7,8 +7,8 @@ Methods to strip repeated elements from a sequence or collection.
 
 The `uniqued()` method returns an array, dropping duplicate elements
 from a sequence. The `uniqued(on:)` method does the same, using 
-the result of the given closure to determine the "uniqueness" of each 
-element.
+the result of the given closure, or a key path, to determine the
+"uniqueness" of each element.
 
 ```swift
 let numbers = [1, 2, 3, 3, 2, 3, 3, 2, 2, 2, 1]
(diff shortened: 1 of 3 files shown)


xwu:
> ```
> let b = a.uniqued(on: { $0.first })
>
> let c = a.uniqued(on: \.first)
> ```

Unless I'm mistaken, this should already work without any additional overloads; that was the purpose of [SE-0249](https://github.com/apple/swift-evolution/blob/main/proposals/0249-key-path-literal-function-expressions.md). If it doesn't work, then we should look into what it would take to make it work without requiring a manually created overload.

PR #85, opened 2021-02-22
[Source Break] Making all collection/sequence wrappers base properties internal 

### LINKED ISSUE #84: Make `base` collections internal
Several of the sequence/collection wrappers provide `public` access to the collection that they wrap, which can end up raising questions about substitutability and isn't strictly necessary. We should make all of these `internal`, instead.

LucianoPAlmeida:
Making all collection/sequence wrappers base properties internal 

Fixes #84 

cc @natecook1000 @timvermeulen 
### Checklist

### DIFF
--- a/Sources/Algorithms/Windows.swift
+++ b/Sources/Algorithms/Windows.swift
@@ -47,8 +47,11 @@ extension Collection {
 /// A collection wrapper that presents a sliding window over the elements of
 /// a collection.
 public struct Windows<Base: Collection> {
-  public let base: Base
-  public let size: Int
+  @usableFromInline
+  internal let base: Base
+  
+  @usableFromInline
+  internal let size: Int
   
   @usableFromInline
   internal var firstUpperBound: Base.Index?
--- a/Sources/Algorithms/Chain.swift
+++ b/Sources/Algorithms/Chain.swift
@@ -14,10 +14,12 @@ public struct Chain2<Base1: Sequence, Base2: Sequence>
   where Base1.Element == Base2.Element
 {
   /// The first sequence in this chain.
-  public let base1: Base1
+  @usableFromInline
+  internal let base1: Base1
   
   /// The second sequence in this chain.
-  public let base2: Base2
+  @usableFromInline
+  internal let base2: Base2
 
   @usableFromInline
   internal init(base1: Base1, base2: Base2) {
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -13,7 +13,8 @@
 /// predicate or projection.
 public struct LazyChunked<Base: Collection, Subject> {
   /// The collection that this instance provides a view onto.
-  public let base: Base
+  @usableFromInline
+  internal let base: Base
   
   /// The projection function.
   @usableFromInline
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -12,7 +12,8 @@
 /// A collection wrapper that generates combinations of a base collection.
 public struct Combinations<Base: Collection> {
   /// The collection to iterate over for combinations.
-  public let base: Base
+  @usableFromInline
+  internal let base: Base
   
   @usableFromInline
   internal let baseCount: Int
--- a/Sources/Algorithms/Cycle.swift
+++ b/Sources/Algorithms/Cycle.swift
@@ -12,7 +12,8 @@
 /// A collection wrapper that repeats the elements of a base collection.
 public struct Cycle<Base: Collection> {
   /// The collection to repeat.
-  public let base: Base
+  @usableFromInline
+  internal let base: Base
   
   @usableFromInline
   internal init(base: Base) {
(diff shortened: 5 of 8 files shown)


Sources/Algorithms/Intersperse.swift
```diff
@@ -12,8 +12,12 @@
 /// A sequence that presents the elements of a base sequence of elements
 /// with a separator between each of those elements.
 public struct Intersperse<Base: Sequence> {
-  let base: Base
-  let separator: Base.Element
+  @usableFromInline
+  internal let base: Base
+  
+  @usableFromInline
+  internal let separator: Base.Element
```
xwu, line 19: Does this _need_ to be made internal? Is there any benefit to making it visible?

Sources/Algorithms/Windows.swift
```diff
@@ -46,8 +46,11 @@ extension Collection {
 /// A collection wrapper that presents a sliding window over the elements of
 /// a collection.
 public struct Windows<Base: Collection> {
-  public let base: Base
-  public let size: Int
+  @usableFromInline
+  internal let base: Base
+  
+  @usableFromInline
+  internal let size: Int
```
xwu, line 53: Ditto here: does `size` need to be made internal?

Sources/Algorithms/Intersperse.swift
LucianoPAlmeida: This is a really good question, the only problem I can think is where `Base.Element` is a reference or we make `set` public, exposing it may allow users to modify state causing the wrapper collection to change behavior, but I'm not sure if this is really a problem ...
 
> Is there any benefit to making it visible?
I can't think of an use case where someone could need access to this property.

xwu: You can't set a stored `let`.

LucianoPAlmeida: Ah! indeed you right, so the only concern would be references then e.g. 
```
struct CollectionWrapper<Base> where Base: Collection {
  let base: Base
  let separator: Base.Element
}

class Ref: Equatable {
  var b: Bool = true
  
  init() {}
  init(_ b: Bool) { self.b = b }
  
  static func == (lhs: Ref, rhs: Ref) -> Bool {
    return lhs.b == rhs.b
  }
}

let l: [Ref] = [Ref(), Ref(false), Ref()]

var c = CollectionWrapper(base: l, separator: Ref(false))

print(c.separator.b)

c.separator.b = true
// Any operation on `c` after that may have behavior changed 

print(c.separator.b) 
```

Although is very specific, this would be possible if we have the public property and can lead to unexpected behavior
But I still not sure if we should consider this really a problem here ... WDYT?

xwu: I'm not sure I see the problem. The same behavior would happen if the reference were mutated not via `c.separator`: that's just what it means to have a reference type.

LucianoPAlmeida: > I'm not sure I see the problem. The same behavior would happen if the reference were mutated not via c.separator: that's just what it means to have a reference type.

Ah, indeed! You are totally right :) 

> Overall, while I don't have a concrete use case in mind, I do rather think it's reasonable for a user to want to know what separator they are using here, and to query that through the type that's using the separator. It is a property that's taken into account as part of the type's equivalence relation, is it not?

I agree this is reasonable and there isn't really a real reason to make it internal, so I think we could change.

timvermeulen: > Overall, while I don't have a concrete use case in mind, I do rather think it's reasonable for a user to want to know what separator they are using here, and to query that through the type that's using the separator.

@xwu Would you say the same argument could be made when deciding whether to make the `base` collection private, and if not, what makes it different?

xwu: The rationale for removing the public accessibility of `base` is given in [issue 84](https://github.com/apple/swift-algorithms/issues/84).

Sources/Algorithms/Windows.swift
xwu: @LucianoPAlmeida I take it the 👍 indicates that you'd agree with reverting this change?

Sources/Algorithms/Intersperse.swift
timvermeulen: Sure, I was just curious whether you thought your reasoning would also extend to the `base` property if we ignore the concern about substitutability. Do you think it's reasonable for a user to want to access the base collection?

Sources/Algorithms/Windows.swift
LucianoPAlmeida: Ah sorry, I just thought that it was the same topic as the on going discussion above... that's why I just reacted and didn't respond here :)

Sources/Algorithms/Intersperse.swift
LucianoPAlmeida: I was thinking 2 things this after @timvermeulen comments 

* Does it make sense to be able to access what separator or size the wrapper collection is using if they can't access the base collection? 
* I was wondering that for the user perspective the most important thing is that this is a collection that behave as expected, kind of in the same sense that an eager implementation that just produces a result array. So by this reasoning specific properties of the wrapper type and I'm not sure but maybe even the wrapper type itself are kind of transparent for the user in the sense that they probably won't even expect to have access to those kind of specific collection properties. This is more a question than an affirmation, I'm still thinking about it... WDYT? @xwu @timvermeulen

xwu: > Sure, I was just curious whether you thought your reasoning would also extend to the base property if we ignore the concern about substitutability. Do you think it's reasonable for a user to want to access the base collection?

My _only_ concern is about substitutability.

natecook1000:
@swift-ci Please test

PR #90, opened 2021-03-07
Simultaneous Minimum and Maximum Evaluation

CTMacUser:
### Description

Adds a method to `Sequence` to evaluate `min(by:)` and `max(by:)` in a single pass.  Adds an overload when the `Element` type is `Comparable` to default the predicate to `<`.

Should fix Swift bug [SR-890](https://bugs.swift.org/browse/SR-890).

### Detailed Design
Include any additional information about the design here. At minimum, show any new API:

```swift
extension Sequence {
    /// Returns both the minimum and maximum elements in the sequence, using
    /// the given predicate as the comparison between elements.
    public func extrema(by areInIncreasingOrder: (Element, Element) throws -> Bool) rethrows -> (min: Element, max: Element)?
}

    /// Returns both the minimum and maximum elements in the sequence.
    @inlinable public func extrema() -> (min: Element, max: Element)?
}
```

### Documentation Plan

There is documentation on the new methods.  Plus the "Minmax" documentation file has been extended with descriptions of these methods.

### Test Plan

The "Minmax" test file has new tests for the new methods.

### Source Impact

The change should be additive.

### Checklist


CTMacUser:
I just splatted my documentation next to the existing docs in that file.  There should be more work to merge them together.

I'm not familiar with Python.  But if their containers have an equivalent method for single-pass min and max, please suggest something to add to the end of the documentation.

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,92 @@ extension Collection where Element: Comparable {
...
+  ///         Optional((key: "Coral", value: 16)) Optional((key: "Heliotrope", value: 296))
+  ///         """
+  ///      */
+  ///
+  /// - Parameter areInIncreasingOrder: A predicate that returns `true`
+  ///   if its first argument should be ordered before its second
+  ///   argument; otherwise, `false`.
+  /// - Returns: A tuple with the sequence's minimum element, followed by its
+  ///   maximum element. For either member, if the sequence provides multiple
+  ///   qualifying elements, the one chosen is unspecified. The same element may
```
natecook1000, line 428: I’d like to take a more principled stand than this. Right now we have two different behaviors for which maximum element is chosen. First, `sort()`, `min(count:)`, and `Swift.max` (not the sequence version) all either return the last equal value or, in the sort/min(count:) case, put it at the end. 

```swift
/// All instances of `Foo` are equal to each other.
class Foo: Comparable {
    static func ==(lhs: Foo, rhs: Foo) -> Bool { true }
    static func < (lhs: Foo, rhs: Foo) -> Bool { false }
}

let a = Foo()
let b = Foo()
max(a, b) === b            // true
[a, b].sorted().last === b // true
```

Second, the sequence version of `max()` returns the _first_ instance of the maximum value.

```swift
[a, b].max() === b         // false
```

On the `min` side, there’s only one behavior — all the equivalent methods return the first minimum element.

Even though we’ll end up where this method and the stdlib’s `Sequence.max()` return different instances, let’s align this new method with the majoritarian position, and I’ll keep looking at how we can switch `Sequence.max()` to come into alignment.

natecook1000 (commented):
Thanks for this addition, @CTMacUser! 📉📈 

I’m still mulling over the method name. It looks like there’s some good input [on the forums](https://forums.swift.org/t/simultaneous-min-max/45414) — thanks for starting that thread! I have a couple more notes below.

Sources/Algorithms/MinMax.swift
CTMacUser: Enshrining which element is chosen to break ties seems like an over-specification; someone's code may be buggy if they're depending on a specific equivalent element.

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,92 @@ extension Collection where Element: Comparable {
...
+    var iterator = makeIterator()
+    guard var lowest = iterator.next() else { return nil }
+
+    var highest = lowest
+    while let element = iterator.next() {
+      if try areInIncreasingOrder(element, lowest) {
+        lowest = element
+      } else if try areInIncreasingOrder(highest, element) {
+        highest = element
+      }
```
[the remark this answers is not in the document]
CTMacUser: I'm not familiar with this optimization.  Is it documented anywhere?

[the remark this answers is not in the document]
natecook1000: It's described in more detail in [the original bug](https://bugs.swift.org/browse/SR-890) — please let me know if you need more info!

Sources/Algorithms/MinMax.swift
natecook1000: We already specify this in e.g. `Swift.max`, and I'd like to make the `sort()` stability a guarantee, which is similar. The dependence is exactly why I'd like to specify the behavior — rather than someone depending on the result of an implementation quirk, they can depend on documented behavior, and we can test to make sure we meet the expectation.

CTMacUser:
I conditionally added @natecook1000 's suggestion to test two elements at a time within the loop.  Can someone actually measure this with both short and (very?) long sequences?  It looks like the sort of slick programming you'd think improves runtimes, but actually doesn't (or does so rarely it's not worth it).

natecook1000:
@CTMacUser I did some benchmarking and got I think what I'd expect — when the comparison is fast, there isn't a big difference between the two implementations, but when the comparison is non-trivial, the optimized version wins, as it's only performing 3/4 the number of comparisons.

Here's a comparison of running `min()` and `max()` separately, along with the two `minAndMax()` implementations, on an array of `Double`s. Comparing doubles is fast, so there's not much difference between the two single-pass versions. The primary savings here is in skipping the second pass:

doubles

When we compare strings with matching prefixes, on the other hand, the comparison time swamps the extra iteration, and the optimized version wins out:

strings

natecook1000:
@CTMacUser I did some benchmarking and got I think what I'd expect — when the comparison is fast, there isn't a big difference between the two implementations, but when the comparison is non-trivial, the optimized version wins, as it's only performing 3/4 the number of comparisons.

Here's a comparison of running `min()` and `max()` separately, along with the two `minAndMax()` implementations, on an array of `Double`s. Comparing doubles is fast, so there's not much difference between the two single-pass versions. The primary savings here is in skipping the second pass:

doubles

When we compare strings with matching prefixes, on the other hand, the comparison time swamps the extra iteration, and the optimized version wins out:

strings

natecook1000:
@CTMacUser This looks ready to go! Want to remove the conditional bits of the implementation and we can merge?

Guides/MinMax.md
```diff
@@ -39,20 +60,26 @@ extension Collection where Element: Comparable {
...
+
+extension Sequence where Element: Comparable {
+    public func minAndMax() -> (min: Element, max: Element)?
+}
 ```
 
 ### Complexity
 
-The algorithm used is based on [Soroush Khanlou's research on this matter](https://khanlou.com/2018/12/analyzing-complexity/). The total complexity is `O(k log k + nk)`, which will result in a runtime close to `O(n)` if *k* is a small amount. If *k* is a large amount (more than 10% of the collection), we fall back to sorting the entire array. Realistically, this means the worst case is actually `O(n log n)`.
+The algorithm used for an extrema block is based on [Soroush Khanlou's research on this matter](https://khanlou.com/2018/12/analyzing-complexity/). The total complexity is `O(k log k + nk)`, which will result in a runtime close to `O(n)` if *k* is a small amount. If *k* is a large amount (more than 10% of the collection), we fall back to sorting the entire array. Realistically, this means the worst case is actually `O(n log n)`.
```
xwu, line 71: What is an "extrema block"? Can you use other terminology that would be more generally familiar to users?

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,129 @@ extension Collection where Element: Comparable {
...
+  ///   (Transitive incomparability)
+  ///
+  /// This example shows how to use the `minAndMax(by:)` method on a
+  /// dictionary to find the key-value pair with the lowest value and the pair
+  /// with the highest value.
+  ///
+  ///     let hues = ["Heliotrope": 296, "Coral": 16, "Aquamarine": 156]
+  ///     let minmax = hues.minAndMax { a, b in a.value < b.value }
+  ///     let leastHue = minmax?.min, greatestHue = minmax?.max
+  ///     print(leastHue, greatestHue)
```
xwu, line 415: Nit: can we simplify this code example by force unwrapping? The user doesn't need to read `Optional(...)` printed twice to understand the concept here, so best not to clutter. Something like (typed freehand--please check it for accuracy):

```
print(minmax!)
// Prints "(min: (key: "Coral", value: 16), max: (key: "Heliotrope", value: 296))"
```

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,129 @@ extension Collection where Element: Comparable {
...
+      } else {
+        if try areInIncreasingOrder(element, lowest) {
+          lowest = element
+        }
+        if try !areInIncreasingOrder(following, highest) {
+          highest = following
+        }
+      }
+    }
+    #else
```
xwu, line 472: 🤔

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,117 @@ extension Collection where Element: Comparable {
...
+        if try areInIncreasingOrder(element, lowest) {
+          lowest = element
+        }
+        if try !areInIncreasingOrder(following, highest) {
+          highest = following
+        }
+      }
+    }
+    return (lowest, highest)
+  }
```
NevinBR, line 469: The while loop can be substantially simplified, since iterators are guaranteed to keep returning `nil` after they’ve done so once:

```swift
while var low = iterator.next() {
  var high = iterator.next() ?? low
  if try areInIncreasingOrder(high, low) { swap(&low, &high) }
  if try areInIncreasingOrder(low, lowest) { lowest = low }
  if try !areInIncreasingOrder(high, highest) { highest = high }
}
```

This does technically perform one extraneous comparison on odd-length sequences (`high` vs `low` after the `??` was triggered), but even that can be rectified by restoring the `guard` if benchmarks show a measurable difference:

```swift
while var low = iterator.next() {
  guard var high = iterator.next() else {
    if try areInIncreasingOrder(low, lowest) { lowest = low }
    if try !areInIncreasingOrder(low, highest) { highest = low }
    break
  }
  if try areInIncreasingOrder(high, low) { swap(&low, &high) }
  if try areInIncreasingOrder(low, lowest) { lowest = low }
  if try !areInIncreasingOrder(high, highest) { highest = high }
}
```

NevinBR:
Has a 4-at-a-time version of `minAndMax` been benchmarked?

It doesn’t change the total number of comparisons versus the pairwise version, but per [Steve Canon’s remarks](https://forums.swift.org/t/simultaneous-min-max/45414/20) on the forums it may still run faster by halving the loop-carried dependencies.

And it’s pretty easy to write:

```swift
while let a = iterator.next() {
  let b = iterator.next() ?? a
  let c = iterator.next() ?? b
  let d = iterator.next() ?? c
  let (low, high) = minAndMaxOf4(a, b, c, d)
  if try areInIncreasingOrder(low, lowest) { lowest = low }
  if try !areInIncreasingOrder(high, highest) { highest = high }
}
```

Using the helper function:

```swift
func minAndMaxOf4<T>(
  _ a: T, _ b: T, _ c: T, _ d: T,
  by areInIncreasingOrder: (T, T) throws -> Bool
) rethrows -> (min: T, max: T) {
  var (a, b, c, d) = (a, b, c, d)
  try sort(&a, &b, by: areInIncreasingOrder)
  try sort(&c, &d, by: areInIncreasingOrder)
  try sort(&a, &c, by: areInIncreasingOrder)
  try sort(&b, &d, by: areInIncreasingOrder)
  return (a, d)
}

func sort<T>(
  _ a: inout T, _ b: inout T,
  by areInIncreasingOrder: (T, T) throws -> Bool
) rethrows {
  if try areInIncreasingOrder(b, a) { swap(&a, &b) }
}
```

CTMacUser:
In my quick runs of the already-provided tests, using the four-at-a-time loop results in more calls to the comparison predicate than the improved two-at-a-time loop did (78 vs. 68).

NevinBR:
> In my quick runs of the already-provided tests, using the four-at-a-time loop results in more calls to the comparison predicate than the improved two-at-a-time loop did (78 vs. 68).

(a) There should be at most 4 extra comparisons the way I wrote it, and they can be avoided by replacing the `??`s with `if let`s.

(b) The important thing to compare is the actual time the algorithm takes to run.

(c) The potential time-save comes from reducing the loop-carried dependencies, not the number of comparisons.

Guides/MinMax.md
```diff
@@ -39,20 +60,26 @@ extension Collection where Element: Comparable {
...
 
 ![Benchmark](Resources/SortedPrefix/FewElements.png)
 ![Benchmark 2](Resources/SortedPrefix/ManyElements.png)
 
+The algorithm used for simultaneous minimum and maximum is slightly optimized. At each iteration, two elements are read, their relative order is determined, then each is compared against exactly one of the current extrema for potential replacement. When a comparison predicate has to analyze every component of both operands, the optimized algorithm isn't much faster than the straightforward approach. But when a predicate only needs to compare a small part of each instance, the optimization shines through.
+
 ### Comparison with other languages
 
-**C++:** The `<algorithm>` library defines a `partial_sort` function where the entire array is returned using a partial heap sort.
+**C++:** The `<algorithm>` library defines a `partial_sort` function where the entire array is returned using a partial heap sort. It also defines a `minmax_element` function that scans a range for its minimal and maximal elements.
```
LemonSpike, line 82: Nit: Maybe reword to `scans a range for its minimum and maximum values.`?

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,141 @@ extension Collection where Element: Comparable {
...
+    ) throws -> (min: Element, max: Element) {
+      var (a, b, c, d) = (a, b, c, d)
+      try sort(&a, &b)
+      try sort(&c, &d)
+      try sort(&a, &c)
+      try sort(&b, &d)
+      return (a, d)
+    }
+
+    // Read the elements in four-at-a-time.  Some say this is more effective
```
LemonSpike, line 481: Nit: Double spaces.

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,141 @@ extension Collection where Element: Comparable {
...
+    // Read the elements in pairwise.  Structuring the comparisons around this
+    // is actually faster than loops based on extracting and testing elements
+    // one-at-a-time.
+    while var low = iterator.next() {
+      if var high = iterator.next() {
+        // Update the upper bound with the larger new element.
+        if try areInIncreasingOrder(high, low) { swap(&low, &high) }
+        if try !areInIncreasingOrder(high, highest) { highest = high }
+      } else {
+        // Update the upper bound by reusing the last element.  The next element
```
LemonSpike, line 453: Are double spaces standard for `//`? I see that the `///` comments have triple spaces but sometimes double slash comments have double, just wanted to check (obviously very minor! 😊).

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,141 @@ extension Collection where Element: Comparable {
...
+  ///   according to the predicate. If `a` and `b` are incomparable, and `b`
+  ///   and `c` are incomparable, then `a` and `c` are also incomparable.
+  ///   (Transitive incomparability)
+  ///
+  /// This example shows how to use the `minAndMax(by:)` method on a
+  /// dictionary to find the key-value pair with the lowest value and the pair
+  /// with the highest value.
+  ///
+  ///     let hues = ["Heliotrope": 296, "Coral": 16, "Aquamarine": 156]
+  ///     if let extremeHues = hues.minAndMax(by: {$0.value < $1.value}) {
```
LemonSpike, line 413: Nit: Spaces between the curly braces.

Sources/Algorithms/MinMax.swift
```diff
@@ -383,3 +383,141 @@ extension Collection where Element: Comparable {
...
+  /// - Returns: A tuple with the sequence's minimum element, followed by its
+  ///   maximum element. For either member, if there is a tie for the extreme
+  ///   value, the element chosen is unspecified. The same element may be used
+  ///   for both members if all the elements are equal. If the sequence has no
+  ///   elements, returns `nil`.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the sequence.
+  @inlinable
+  public func minAndMax() -> (min: Element, max: Element)? {
+    return minAndMax(by: <)
```
LemonSpike, line 521: Nit: We don't need a return here.

NevinBR: I like the `return`.

As a matter of style, I think `return` should only be omitted when the entire`{...}` block is on a single line.

LemonSpike: I wish we had a style guide for these things.

natecook1000:
@CTMacUser My benchmarking shows very small differences between find the min and max two-at-a-time vs four-at-a-time; which one wins depends on the cost of the comparison. In any case, we don't need to let that decision block merging this addition — could you choose one implementation and we'll get this merged? Thanks!

NevinBR:
> @CTMacUser My benchmarking shows very small differences between find the min and max two-at-a-time vs four-at-a-time; which one wins depends on the cost of the comparison. In any case, we don't need to let that decision block merging this addition — could you choose one implementation and we'll get this merged? Thanks!

Thanks for testing, Nate. If the performance is similar, then pairwise is clearly preferable for implementation simplicity.

natecook1000:
@swift-ci Please test

natecook1000:
@CTMacUser I can single out the implementation post-merge — thanks again for your contribution! 👏🏻👏🏻

CTMacUser:
Sorry for flaking out, but it seems everything got handled anyway.  I've spent the time procuring an ARM-based Mac and transitioning my programming to it.

natecook1000:
No worries, and congrats! 🎉

PR #92, opened 2021-03-09
Internally add Collection.endOfPrefix(while:) and BidirectionalCollection.startOfSuffix(while:)

timvermeulen:
Internally adds the `endOfPrefix(while:)` method to `Collection` and the `startOfSuffix(while:)` method to `BidirectionalCollection`.

These methods return the same values as `prefix(while: predicate).endIndex` and `suffix(while: predicate).startIndex`, respectively. We're adding them as internal methods for now because we're already using this functionality in several places throughout the package, but it's unclear whether these are the names we want to stick with.

### Checklist

### DIFF
--- a/Sources/Algorithms/Suffix.swift
+++ b/Sources/Algorithms/Suffix.swift
@@ -37,3 +37,59 @@ extension BidirectionalCollection {
     return self[result...]
   }
 }
+
+//===----------------------------------------------------------------------===//
+// endOfPrefix(while:)
+//===----------------------------------------------------------------------===//
+
+extension Collection {
+  /// Returns the exclusive upper bound of the prefix of elements that satisfy
+  /// the predicate.
+  ///
+  /// - Parameter predicate: A closure that takes an element of the collection
+  ///   as its argument and returns `true` if the element is part of the prefix
+  ///   or `false` if it is not. Once the predicate returns `false` it will not
+  ///   be called again.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the collection.
+  @usableFromInline
+  internal func endOfPrefix(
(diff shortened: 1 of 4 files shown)


Sources/Algorithms/Trim.swift
```diff
@@ -29,20 +29,8 @@ extension BidirectionalCollection {
...
-        return self[sliceStart..<sliceEnd]
-      }
-      sliceEnd = idxBeforeSliceEnd
-    }
-    
-    // Trimmed everything.
-    return self[Range(uncheckedBounds: (sliceStart, sliceStart))]
+    let start = try endOfPrefix(while: predicate)
+    let end = try self[start...].startOfSuffix(while: predicate)
+    return self[start..<end]
```
natecook1000, line 34: This is amazing ❤️

natecook1000 (approved):
LGTM!

Sources/Algorithms/Suffix.swift
```diff
@@ -27,13 +27,62 @@ extension BidirectionalCollection {
...
+  /// Returns the inclusive lower bound of the suffix of elements that satisfy
+  /// the predicate.
+  ///
+  /// - Parameter predicate: A closure that takes an element of the collection
+  ///   as its argument and returns `true` if the element is part of the suffix
+  ///   or `false` if it is not. Once the predicate returns `false` it will not
+  ///   be called again.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the collection.
+  @usableFromInline
```
karwa, line 74: This PR looks really nice, but I think these should be inlinable rather than just usableFromInline. We want the compiler to be able to specialise them for a particular predicate closure.

PR #93, opened 2021-03-09
Update swift-tools-version to 5.3 :package:

LemonSpike:
When running this package from the command line with `swift build` I got the warning:

> error: package at '<insert file path here>' is using Swift tools version 3.1.0 which is no longer supported; consider using '// swift-tools-version:5.3' to specify the current tools version

Hence, this is a minor change that I built on my local main in Xcode with the `swift-tools-version:5.3` specified and there seemed to be no breaking changes. I don't believe adding/modifying tests would be necessary for tooling rather than code changes, but please let me know if I need to change anything.

### Checklist
- [x] I've added at least one test that validates that my change is working, if appropriate (N/A)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift Algorithms open source project


natecook1000:
Hi @LemonSpike — I'm a bit confused by this error, since the Package.swift file should be at version 5.1, not 3.1. Can you provide more information about how you're seeing this error, along with your versions of macOS and Xcode?

natecook1000:
It sounds like SwiftPM isn't recognizing or seeing the `swift-tools-version` line in your Package.swift file, not the one that's part of the Algorithms package. Can you check that it's present and formatted correctly?

LemonSpike:
Ahh oops. I fixed my `Package.swift` and everything magically works now ™. Thanks for the help!

PR #103, opened 2021-03-13
Add a breadth-first search algorithm over any tree

mdznr:
### Description
Adds a breadth-first search algorithm that can be applied to any tree structure. This is useful for traversing tree data structures that appear all over our apps, like the hierarchies for views, layers, view controllers, windows, and more.

### Detailed Design
Traverse a tree in breadth-first order by providing the root of the tree, and a closure that returns the children of a given node.

```swift
let tree = Node(element: 1, children: [
	Node(element: 2, children: [
		Node(element: 5, children: [
			Node(element: 9),
			Node(element: 10),
		]),
		Node(element: 6),
	]),
	Node(element: 3, children: [
	]),
	Node(element: 4, children: [
		Node(element: 7, children: [
			Node(element: 11),
			Node(element: 12),
		]),
		Node(element: 8),
	]),
])

let elements = BreadthFirstTreeSequence(root: tree, children: { $0.children })
// Array(elements) == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
```

Breadth-first-tree

### Open Questions
- [ ] **API Design and Future Considerations**
I’d like to consider how other tree traversal algorithms—namely depth-first search—would work and see how that may influence the API and implementation. Would there be shared code between the two?

- [ ] **Naming**
The majority of the additions in swift-algorithms are functions based on `Sequence`s and `Collection`s. Since there is no formal tree data structure in Swift, there isn’t really an existing structure to extend, thus the creation of a new structure. The structure represents a tree flattened into a sequence in breadth-first order. I think “breadth-first” should definitely be included in the name, although the term “search” doesn’t always apply. The sequence is great at searching, but it isn’t limited to that. You may want the breadth-first _ordering_ of something without necessarily searching for one element in particular.

- [ ] **Documentation**
I’d love to improve the documentation on how this works, particularly after the API settles down. I’m unsure how much depth to go into in how breadth-first search works as an algorithm versus assuming the reader is familiar with this well-known algorithm.

### Documentation Plan
- Added BreadthFirstSearch.md
- Added `BreadthFirstTreeSequence` to README.md
- Added inline documentation to implementation and unit tests

### Test Plan
Adds many unit tests for:
- Empty tree
- Simple tree
- Complex tree
- The number of times a closure is called during iteration of early termination algorithms, like `first(where:)`

### Source Impact
This is purely additive.

### Checklist

### DIFF
--- a/Guides/BreadthFirstSearch.md
+++ b/Guides/BreadthFirstSearch.md
@@ -1,0 +1,55 @@
+# Breadth First Search
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/BreadthFirstSearch.swift) | 
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/BreadthFirstSearch.swift)]
+
+Traverses a tree structure in breadth-first order.
+
+This traversal can occur on any tree structure by using `BreadthFirstTreeSequence`:
+
+```swift
+let tree = Node(element: "1", children: [
+	Node(element: "2", children: [
+		Node(element: "5", children: [
+			Node(element: "9"),
+			Node(element: "10"),
+		]),
+		Node(element: "6"),
+	]),
+	Node(element: "3"),
+	Node(element: "4", children: [
+		Node(element: "7", children: [
+			Node(element: "11"),
+			Node(element: "12"),
+		]),
+		Node(element: "8"),
+	]),
+])
+
+let elements = BreadthFirstTreeSequence(root: tree, children: { $0.children })
+// Array(elements) == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+```
+
+## Detailed Design
+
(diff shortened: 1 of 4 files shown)


natecook1000:
Thanks for opening this PR, @mdznr! 

This is an area where we need some design exploration about how to express graphs before we get into algorithms like BFS. Is there a design for a `Graph` protocol that would enable this and other algorithms? We may want to think about how and whether we'd want to express weighted vs unweighted graphs, directionality, and other properties of a graph and its nodes.

To start out, we'll need to work up some example algorithms (like this one) and graph implementations (adjacency lists, generated graphs, etc) to find the right patterns before moving on to actual API.

kylemacomber:
In the [announcement blog post](https://swift.org/blog/swift-algorithms/#contribution-criteria) we said:

> The immediate focus of the package is to incubate a pragmatic set of algorithms generalized over the Sequence and Collection family of protocols for eventual inclusion in the Swift standard library

I think it's still too early for the Algorithms package to expand its focus to algorithms over non-linear data structures.

PR #104, opened 2021-03-13
Add `trimmingSuffix`, `trimmingPrefix` and mutating variants

### LINKED ISSUE #101: Add `trimmingPrefix(while:)`, `trimmingSuffix(while:)`, and mutating variants
We currently provide just one way to trim a collection:

```swift
  public func trimming(
  ) rethrows -> SubSequence {
    let start = try endOfPrefix(while: predicate)
    let end = try self[start...].startOfSuffix(while: predicate)
    return self[start..<end]
  }
}
```

We should also provide ways to trim only the start of the collection (`trimmingPrefix(while:)`, available to all collections), or only the end (`trimmingSuffix(while:)`).

Furthermore, collections for which `Self == SubSequence` should have `mutating` variants using `trim` as the base name. As an example, `Collection`'s [`popFirst()`](https://github.com/apple/swift/blob/main/stdlib/public/core/Collection.swift#L1047-L1062) from the standard library is implemented using the same constraint.

And finally, despite `String` not being its own `SubSequence`, `String` should also have access to these mutating variants for the sake of convenience.

fedeci:
### Description
Closes #101 
`trimming(while:)` is split in two methods `trimmingPrefix(while:)` which is available to all `Collection`s and `trimmingSuffix(while:)` available only to `BidirectionalCollection`s.
This also adds a mutable variant of each `trim` method where `SubSequence == Self`.

### Detailed Design

```swift
func trimmingSuffix(
) rethrows -> SubSequence {
  let end = try self[startIndex...].startOfSuffix(while: predicate)
  return self[startIndex..<end]
}

func trimmingPrefix(
) rethrows -> SubSequence {
  let start = try endOfPrefix(while: predicate)
  return self[start..<endIndex]
}

func trimming(
) rethrows -> SubSequence {
  return try trimmingPrefix(while: predicate).trimmingSuffix(while: predicate)
}
```
The same applies to mutable `trim`s.
- [x] `String` must access mutable `trim`s

### Documentation Plan
Docs not yet updated.

### Test Plan
Since the two new methods are just a refactoring of `trimming` all those tests also apply to the new API.
- [x] Mutating variants tests added.

### Source Impact
It doesn't break anything.

### Checklist

### DIFF
--- a/Guides/Trim.md
+++ b/Guides/Trim.md
@@ -48,6 +48,25 @@ func myAlgorithm2<Input>(input: Input) where Input: BidirectionalCollection {
 Swift provides the `BidirectionalCollection` protocol for marking types which support reverse traversal,
 and generic types and algorithms which want to make use of that should add it to their constraints.
 
+### >= 0.3.0
+
+In `v0.3.0` new methods are added to allow discarding all the elements matching the predicate at the beginning (prefix) or at the ending (suffix) of the collection.
+- `trimmingSuffix(while:)` can only be run on collections conforming to the `BidirectionalCollection` protocol.  
+- `trimmingPrefix(while:)` can be run also on collections conforming to the `Collection` protocol.
+
+```swift
+let myString = "   hello, world  "
+print(myString.trimmingPrefix(while: \.isWhitespace)) // "hello, world   "
+
+print(myString.trimmingSuffix(while: \.isWhitespace)) // "   hello, world"
+```
+Also mutating variants for all the methods already existing and the new ones are added.
+```swift
+var myString = "   hello, world  "
+myString.trim(while: \.isWhitespace)
+print(myString) // "hello, world"
+```
+
 ### Complexity
 
 Calling this method is O(_n_).
--- a/Sources/Algorithms/Trim.swift
+++ b/Sources/Algorithms/Trim.swift
@@ -34,6 +34,15 @@ extension Collection {
   }
 }
 
+extension Collection where Self == SubSequence {
+  @inlinable
+  public mutating func trimPrefix(
+    while predicate: (Element) throws -> Bool
+  ) rethrows {
+    self = try trimmingPrefix(while: predicate)
+  }
+}
+
 extension BidirectionalCollection {
   /// Returns a `SubSequence` formed by discarding all elements at the start and
   /// end of the collection which satisfy the given predicate.
@@ -78,5 +87,21 @@ extension BidirectionalCollection {
   ) rethrows -> SubSequence {
     let end = try self[startIndex...].startOfSuffix(while: predicate)
     return self[startIndex..<end]
+  }
+}
+
+extension BidirectionalCollection where Self == SubSequence {
+  @inlinable
+  public mutating func trim(
+    while predicate: (Element) throws -> Bool
+  ) rethrows {
+    self = try trimming(while: predicate)
+  }
+  
+  @inlinable
+  public mutating func trimSuffix(
+    while predicate: (Element) throws -> Bool
+  ) rethrows {
+    self = try trimmingPrefix(while: predicate)
   }
 }
--- a/Sources/Algorithms/Trim.swift
+++ b/Sources/Algorithms/Trim.swift
@@ -43,6 +43,7 @@ extension Collection {
 
 extension Collection where Self: RangeReplaceableCollection {
   @inlinable
+  @_disfavoredOverload
   public mutating func trimPrefix(
     while predicate: (Element) throws -> Bool
   ) rethrows {
@@ -117,6 +118,7 @@ extension BidirectionalCollection {
 
 extension BidirectionalCollection where Self: RangeReplaceableCollection {
   @inlinable
+  @_disfavoredOverload
   public mutating func trim(
     while predicate: (Element) throws -> Bool
   ) rethrows {
@@ -124,6 +126,7 @@ extension BidirectionalCollection where Self: RangeReplaceableCollection {
   }
   
   @inlinable
+  @_disfavoredOverload
   public mutating func trimSuffix(
     while predicate: (Element) throws -> Bool
   ) rethrows {
--- a/Sources/Algorithms/Trim.swift
+++ b/Sources/Algorithms/Trim.swift
@@ -42,6 +42,22 @@ extension Collection {
 //===----------------------------------------------------------------------===//
 
 extension Collection where Self: RangeReplaceableCollection {
+  /// Mutates a `Collection` by discarding all elements at the start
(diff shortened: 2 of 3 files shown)


Sources/Algorithms/Trim.swift
```diff
@@ -29,8 +63,45 @@ extension BidirectionalCollection {
...
+extension BidirectionalCollection where Self == SubSequence {
+  @inlinable
+  public mutating func trim(
+    while predicate: (Element) throws -> Bool
+  ) rethrows {
+    self = try trimming(while: predicate)
+  }
+  
+  @inlinable
+  public mutating func trimSuffix(
```
fedeci, line 102: May return the trimmed elements?

Tests/SwiftAlgorithmsTests/TrimTests.swift
```diff
@@ -56,4 +56,14 @@ final class TrimTests: XCTestCase {
     let results_one = [2, 10, 12, 15, 20, 100].trimming { $0.isMultiple(of: 2) }
     XCTAssertEqual(results_one, [15])
   }
+
```
LucianoPAlmeida, line 59: Could we add unit test cases for `trim`?

Sources/Algorithms/Trim.swift
```diff
@@ -29,8 +63,45 @@ extension BidirectionalCollection {
...
+    while predicate: (Element) throws -> Bool
+  ) rethrows -> SubSequence {
+    let end = try self[startIndex...].startOfSuffix(while: predicate)
+    return self[startIndex..<end]
+  }
+}
+
+extension BidirectionalCollection where Self == SubSequence {
+  @inlinable
+  public mutating func trim(
```
LucianoPAlmeida, line 95: For what I understood from the issue we should also add an unconstrained version of `trim`, right @timvermeulen ?

Tests/SwiftAlgorithmsTests/TrimTests.swift
fedeci: Yes, I added a checkbox in the PR description. I was just waiting for a review before adding those.

natecook1000 (commented):
Thanks for this, @fedeci! 👏

I have a few notes for you below, and we also need a couple overloads for each of the mutating methods. The current ones are available for `RangeReplaceableCollection` types; we can also do these operations on self-slicing types (e.g. `where Self == Self.SubSequence`). One such type is a `Range<Int>` — when you write `(0..<10).dropFirst(3)` you end up with the range `3..<10`. Unfortunately, there are also self-slicing, range-replaceable types (like `Substring` and `ArraySlice`), so we'll _also_ need overloads that have both constraints.

Sources/Algorithms/Trim.swift
natecook1000: I don't think that's necessary, since (if my math is right) that could end up requiring an extra allocation just to store the discarded elements.

fedeci:
Do `Substring` and `ArraySlice` inherit the mutating methods from `BidirectionalCollection where Self: RangeReplaceableCollection`, don't they?

Sources/Algorithms/Trim.swift
```diff
@@ -29,8 +84,66 @@ extension BidirectionalCollection {
...
+  @inlinable
+  public mutating func trimSuffix(
+    while predicate: (Element) throws -> Bool
+  ) rethrows {
+    let start = try startOfSuffix(while: predicate)
+    removeSubrange(start..<endIndex)
+  }
+}
+
+extension BidirectionalCollection where Self == Self.SubSequence {
```
natecook1000, line 135: These overloads look great! Unfortunately, for collections that meet both requirements (i.e. range-replaceable collections that are also self-slicing), trying to call the method ends up ambiguous. We used to fix this by adding a third overload in an extension with _both_ constraints, so that it would be more specific than the other two, but now we have the `@_disfavoredOverload` attribute. Can you add `@_disfavoredOverload` to the `RangeReplaceableCollection` versions of the methods?

Tests/SwiftAlgorithmsTests/TrimTests.swift
```diff
@@ -56,4 +56,14 @@ final class TrimTests: XCTestCase {
...
+    
+  func testTrimmingPrefix() {
+    let results = [2, 10, 12, 15, 20, 100].trimmingPrefix { $0.isMultiple(of: 2) }
+    XCTAssertEqual(results, [15, 20, 100])
+  }
+
+  func testTrimmingSuffix() {
+    let results = [2, 10, 12, 15, 20, 100].trimmingSuffix { $0.isMultiple(of: 2) }
+    XCTAssertEqual(results, [2, 10, 12, 15])
+  }
```
natecook1000, line 68: We'll also need tests for the new mutating methods. Can you include a test that uses `ArraySlice` as a type to verify that the ambiguity from above is resolved? For example:

```swift
var values = [2, 10, 12, 15, 20, 100] as ArraySlice
values.trim(while: { $0.isMultiple(of: 2) })
```

Sources/Algorithms/Trim.swift
```diff
@@ -7,6 +7,61 @@
...
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// trimPrefix(while:)
+//===----------------------------------------------------------------------===//
+
+extension Collection where Self: RangeReplaceableCollection {
+  @inlinable
+  public mutating func trimPrefix(
```
natecook1000, line 46: Could you add docs for the mutating methods as well?

natecook1000 (commented):
@fedeci Looking good! A few notes below and then we can merge this addition.

fedeci:
Are still missing documents in the `docs/` folder.

natecook1000:
@fedeci This looks ready to go! Would you like to add a skeleton guide and mark this as ready for review?

fedeci:
Sure, just give me a couple of days because I am full of exams in this period.

natecook1000:
@swift-ci Please test

fedeci:
Here it is :)

Guides/Trim.md
```diff
@@ -48,6 +48,25 @@ func myAlgorithm2<Input>(input: Input) where Input: BidirectionalCollection {
 Swift provides the `BidirectionalCollection` protocol for marking types which support reverse traversal,
 and generic types and algorithms which want to make use of that should add it to their constraints.
 
+### >= 0.3.0
```
fedeci, line 51: I added `0.3.0` supposing that this will be released in a next minor version, but I am not sure

natecook1000:
@swift-ci Please test

natecook1000:
🎉 :shipit:

Issue #100, opened 2021-03-12
Change `XCTAssertEqualSequences` to report how the sequences are different


timvermeulen:
Our test suite contains the `XCTAssertEqualSequences` function which tests whether two sequences have equal elements, either based on `Equatable` conformance or using a custom predicate. It's currently implemented in terms of the `elementsEqual` method on `Sequence` which only returns whether or not the two sequences have equal elements.

For our purposes, when the two sequences are different, it would not only be useful to know _that_ they are different, but also _how_ they are different. Either the elements at the same position in the respective sequences aren't equal, or one of the sequences is shorter than the other one. `XCTAssertEqualSequences` should be changed to log this information to make debugging a bit easier.

hashemi:
I'd love to work on it

LucianoPAlmeida:
@timvermeulen PR was merged this can be closed :)

PR #105, opened 2021-03-13
Improve XCTAssertEqualSequences to log difference

hashemi:
Implements improvement suggested @timvermeulen in #100

Improves the `XCTAssertEqualSequences` function to log when two elements in a sequence are not equivalent or when one of the sequences is shorter than the other.

I implemented it such that as soon as it finds to unequal elements between the sequences, it'll fail and stop without looking at the remaining elements. We could try to make it more sophisticated if that's the goal.

I'm not sure how to test it, since it's a test helper utility. As it stands, `XCTAssertEqualSequences` was not being tested. The test suite, which heavily depends on the function still passes after the changes but that doesn't exercise the function in the failure cases or its output. I experimented with that by changing some tests to make them fail and it seemed to work fine.

The code was based on the implementation of `elementsEqual` from the Swift standard library.

### Checklist
- [X] I've followed the code style of the rest of the project
- [X] I've read the [Contribution Guidelines](../CONTRIBUTING.md)
- [X] I've updated the documentation if necessary

### DIFF
--- a/Tests/SwiftAlgorithmsTests/TestUtilities.swift
+++ b/Tests/SwiftAlgorithmsTests/TestUtilities.swift
@@ -65,7 +65,8 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
 ) rethrows where S1.Element == S2.Element {
 
   func fail(_ reason: String) {
-    XCTFail(message().isEmpty ? reason : "\(message()) - \(reason)",
+    let message = message()
+    XCTFail(message.isEmpty ? reason : "\(message) - \(reason)",
             file: file, line: line)
   }
 
--- a/Tests/SwiftAlgorithmsTests/TestUtilities.swift
+++ b/Tests/SwiftAlgorithmsTests/TestUtilities.swift
@@ -78,7 +78,7 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
       idx += 1
       continue
     case let (e1?, e2?):
-      fail("unequal elements at index \(idx) (\(e1) != \(e2))")
+      fail("element \(e1) on first sequence does not match element \(e2) on second sequence at position \(idx)")
     case (_?, nil):
       fail("second sequence shorter than first")
     case (nil, _?):
--- a/Tests/SwiftAlgorithmsTests/TestUtilities.swift
+++ b/Tests/SwiftAlgorithmsTests/TestUtilities.swift
@@ -63,25 +63,30 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
   _ message: @autoclosure () -> String = "",
   file: StaticString = #file, line: UInt = #line
 ) rethrows where S1.Element == S2.Element {
-    var iter1 = try expression1().makeIterator()
-    var iter2 = try expression2().makeIterator()
-    while true {
-      switch (iter1.next(), iter2.next()) {
-      case let (e1?, e2?):
-        XCTAssertTrue(
-          areEquivalent(e1, e2),
-          "Element \(e1) in first sequence not equal to \(e2) in second sequence.",
-          file: file, line: line)
-        return
-      case (_?, nil):
-        XCTFail("Second sequence shorter than first")
-        return
-      case (nil, _?):
-        XCTFail("First sequence shorter than second")
-        return
-      case (nil, nil): return
-      }
+
+  func fail(_ reason: String) {
+    XCTFail(message().isEmpty ? reason : "\(message()) - \(reason)",
+            file: file, line: line)
+  }
+
+  var iter1 = try expression1().makeIterator()
+  var iter2 = try expression2().makeIterator()
+  var idx = 0
+  while true {
+    switch (iter1.next(), iter2.next()) {
+    case let (e1?, e2?) where areEquivalent(e1, e2):
+      idx += 1
+      continue
+    case let (e1?, e2?):
+      fail("unequal elements at index \(idx) (\(e1) != \(e2))")
+    case (_?, nil):
+      fail("second sequence shorter than first")
+    case (nil, _?):
+      fail("first sequence shorter than second")
+    case (nil, nil): break
     }
+    return
+  }
 }
 
 func XCTAssertLazySequence<S: LazySequenceProtocol>(_: S) {}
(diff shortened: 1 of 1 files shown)


Tests/SwiftAlgorithmsTests/TestUtilities.swift
```diff
@@ -62,7 +63,25 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
...
+    var iter1 = try expression1().makeIterator()
+    var iter2 = try expression2().makeIterator()
+    while true {
+      switch (iter1.next(), iter2.next()) {
+      case let (e1?, e2?):
+        XCTAssertTrue(
+          areEquivalent(e1, e2),
+          "Element \(e1) in first sequence not equal to \(e2) in second sequence.",
+          file: file, line: line)
+        return
```
timvermeulen, line 75: This `return` looks out of place here, we only want to return from the other cases.

Tests/SwiftAlgorithmsTests/TestUtilities.swift
```diff
@@ -62,7 +63,25 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
   _ message: @autoclosure () -> String = "",
```
timvermeulen, line 63: We're not currently using this parameter since we're passing custom messages to the assertion functions in the implementation — let's change those to include the `message` parameter.

timvermeulen (commented):
This looks great, @hashemi!

> I implemented it such that as soon as it finds to unequal elements between the sequences, it'll fail and stop without looking at the remaining elements. We could try to make it more sophisticated if that's the goal.

I think this is a sensible approach. Printing more than a few elements could quickly become unwieldy, and seeing the first mismatch should already give the programmer a good idea about what could be going wrong.

> I'm not sure how to test it, since it's a test helper utility. As it stands, `XCTAssertEqualSequences` was not being tested. The test suite, which heavily depends on the function still passes after the changes but that doesn't exercise the function in the failure cases or its output. I experimented with that by changing some tests to make them fail and it seemed to work fine.

Xcode 12.5 will be adding adding [`XCTExpectFailure`](https://developer.apple.com/documentation/xctest/3726077-xctexpectfailure), so we'll be able to properly test this soon 🙂 For now we don't need to worry about this.

What do you think about keeping track of the position in the sequences where the mismatch occurs, and including it in the error message? We can't include the actual indices because the sequences aren't necessarily `Collection`s, but a simple integer counter could already be valuable. For instance, that would make it immediately clear when one sequence is empty and the other one isn't.

Tests/SwiftAlgorithmsTests/TestUtilities.swift
LucianoPAlmeida: I was wondering if doing like this, for two sequences like `[1,1,1,1,1,1,1]` and `[2,2,2,2,2,2,2]` we would show a lot of assertion errors(one for each mismatched element). Maybe a better idea is to collect all the diff on the sequences and show them at the end in a single assertion at the end?

timvermeulen: Oh yeah, good point. I guess we do want to return there, but only when the elements aren't equal — that is, if we stick to the strategy of only reporting the first mismatch.

hashemi: Speaking of testing this function, that was a bug, it was only checking the first element! Ops.

hashemi:
New commit fixes bug and addresses feedback

Tests/SwiftAlgorithmsTests/TestUtilities.swift
```diff
@@ -62,7 +63,30 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
...
+  var iter1 = try expression1().makeIterator()
+  var iter2 = try expression2().makeIterator()
+  var idx = 0
+  while true {
+    switch (iter1.next(), iter2.next()) {
+    case let (e1?, e2?) where areEquivalent(e1, e2):
+      idx += 1
+      continue
+    case let (e1?, e2?):
+      fail("unequal elements at index \(idx) (\(e1) != \(e2))")
```
LucianoPAlmeida, line 81: Just a wording suggestion 
```suggestion
      fail("element \(e1) on first sequence does not match element \(e2) on second sequence at position \(idx)")
```
Is is a bit more verbose but `unequal` here may not be the word since `areEquivalent` is not necessarily equality :)
Also, index could be confusing because a sequence doesn't actually have the concept of indexing ... but also not sure `position` is better

hashemi: Awesome! Thanks Luciano.

Tests/SwiftAlgorithmsTests/TestUtilities.swift
```diff
@@ -62,7 +63,30 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
   _ message: @autoclosure () -> String = "",
   file: StaticString = #file, line: UInt = #line
 ) rethrows where S1.Element == S2.Element {
-  try XCTAssert(expression1().elementsEqual(expression2(), by: areEquivalent), message(), file: file, line: line)
+
+  func fail(_ reason: String) {
+    XCTFail(message().isEmpty ? reason : "\(message()) - \(reason)",
+            file: file, line: line)
+  }
```
timvermeulen, line 70: This looks good! Just to be on the safe side, let's store the message in its own variable first (inside the `fail` function) to avoid calling it twice.

Tests/SwiftAlgorithmsTests/TestUtilities.swift
timvermeulen: Agreed on the "position" wording the avoid confusion with existing index terminology 🙂

Tests/SwiftAlgorithmsTests/TestUtilities.swift
timvermeulen, line 70: This looks good! Just to be on the safe side, let's store the message in its own variable first (inside the fail function) to avoid calling it twice.

hashemi: Done!

kylemacomber (approved):

kylemacomber:
@swift-ci please test

PR #106, opened 2021-03-14
Add new `FiniteCycle` type as the return type of `Collection.cycled(times:)`

### LINKED ISSUE #99: Add new `FiniteCycle` type as the return type of `Collection.cycled(times:)`
`cycled(times:)` is currently implemented as follows:

```swift
extension Collection {
  public func cycled(times: Int) -> FlattenSequence<Repeated<Self>> {
    repeatElement(self, count: times).joined()
  }
}
```

This is suboptimal because `FlattenSequence` does not conform to `RandomAccessCollection` when the base collection does, even though the return value of `cycled(times:)` could! By implementing this method in terms of `FlattenSequence`, we are essentially throwing away the information that each inner collection is the same (and therefore has the same length).

Another way of repeating a collection a fixed number of times is by taking the product of it and an integer range:

```swift
let array = [10, 20, 30]
for (_, x) in product(0..<5, array) {
  print(x)  // 10, 20, 30, 10, 20, 30, 10, 20, 30, 10, 20, 30, 10, 20, 30
}
```

Unlike `FlattenSequence`, the `Product2` collection does conditionally conform to `RandomAccessCollection`.

We should add a new `FiniteCycle` collection as the return type of `product(times:)` that wraps a `Product2<Range<Int>, Base>` value and conforms to `RandomAccessCollection`.

LemonSpike:
This PR aims to fix #99. I have also split up some relevant tests into separate methods for empty collections and changed some tests for `FiniteCycle`. I have also added tuple labels to `Product2` when using the `Iterator.next()` method and also in the `subscript` method. This is a cosmetic improvement, but I felt it may read better.

### Checklist

### DIFF
--- a/Sources/Algorithms/Cycle.swift
+++ b/Sources/Algorithms/Cycle.swift
@@ -110,7 +110,7 @@ extension FiniteCycle: Collection {
 
   @inlinable
   public subscript(_ index: Index) -> Element {
-    product[index.productIndex].element2
+    product[index.productIndex].1
   }
 
   @inlinable
--- a/Sources/Algorithms/Cycle.swift
+++ b/Sources/Algorithms/Cycle.swift
@@ -56,6 +56,47 @@ extension Cycle: Sequence {
 
 extension Cycle: LazySequenceProtocol where Base: LazySequenceProtocol {}
 
+
+/// A collection wrapper that repeats the elements of a base collection for a finite number of times.
+public struct FiniteCycle<Base: Collection> {
+  /// The collection to repeat.
+  public let base: Base
+
+  /// The number of times to repeat the collection.
+  public let times: Int
+
+  @usableFromInline
+  internal init(base: Base, times: Int) {
+    self.base = base
+    self.times = times
+  }
+}
+
+extension FiniteCycle: Sequence {
+  /// The iterator for a `FiniteCycle` sequence.
+  public struct Iterator : IteratorProtocol {
+    @usableFromInline
+    var productIterator: Product2<Range<Int>, Base>.Iterator
+
+    @usableFromInline
+    internal init(_ product: Product2<Range<Int>, Base>) {
+      self.productIterator = product.makeIterator()
+    }
+
+    @inlinable
+    public mutating func next() -> Base.Element? {
+      self.productIterator.next()?.1
+    }
+  }
+
+  public func makeIterator() -> Iterator {
+    let iterations: Range<Int> = 0..<times
+    return Iterator(Product2(iterations, base))
+  }
+}
+
+extension FiniteCycle: LazySequenceProtocol where Base: LazySequenceProtocol {}
+
 //===----------------------------------------------------------------------===//
 // cycled()
 //===----------------------------------------------------------------------===//
--- a/Sources/Algorithms/Cycle.swift
+++ b/Sources/Algorithms/Cycle.swift
@@ -69,7 +69,7 @@ public struct FiniteCycle<Base: Collection> {
   }
 }
 
-extension FiniteCycle: Sequence {
+extension FiniteCycle {
   /// The iterator for a `FiniteCycle` sequence.
   public struct Iterator : IteratorProtocol {
     @usableFromInline
@@ -92,7 +92,8 @@ extension FiniteCycle: Sequence {
   }
 }
 
-extension FiniteCycle: LazySequenceProtocol where Base: LazySequenceProtocol {}
+extension FiniteCycle: LazySequenceProtocol, LazyCollectionProtocol
+  where Base: LazyCollectionProtocol { }
 
 extension FiniteCycle: Collection {
 
--- a/Sources/Algorithms/Cycle.swift
+++ b/Sources/Algorithms/Cycle.swift
@@ -57,7 +57,8 @@ extension Cycle: Sequence {
 extension Cycle: LazySequenceProtocol where Base: LazySequenceProtocol {}
 
 
-/// A collection wrapper that repeats the elements of a base collection for a finite number of times.
+/// A collection wrapper that repeats the elements of a base collection for a
+/// finite number of times.
 public struct FiniteCycle<Base: Collection> {
   /// The collection to repeat.
   public let base: Base
--- a/Sources/Algorithms/Cycle.swift
+++ b/Sources/Algorithms/Cycle.swift
@@ -93,6 +93,53 @@ extension FiniteCycle: Sequence {
 
 extension FiniteCycle: LazySequenceProtocol where Base: LazySequenceProtocol {}
 
+extension FiniteCycle: Collection {
+
+  public typealias Index = Product2<Range<Int>, Base>.Index
+
+  public subscript(_ index: Index) -> Element {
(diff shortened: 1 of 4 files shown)


Sources/Algorithms/Cycle.swift
```diff
@@ -56,6 +56,47 @@ extension Cycle: Sequence {
 
 extension Cycle: LazySequenceProtocol where Base: LazySequenceProtocol {}
 
+
+/// A collection wrapper that repeats the elements of a base collection for a finite number of times.
```
xwu, line 60: Nit: Please wrap to 80 columns, here and elsewhere.

Sources/Algorithms/Cycle.swift
```diff
@@ -56,6 +56,47 @@ extension Cycle: Sequence {
...
+public struct FiniteCycle<Base: Collection> {
+  /// The collection to repeat.
+  public let base: Base
```
xwu, line 63: Let's make this `internal`, as well as `times`.

This is because equality (we could conform this type to `Equatable`, but it doesn't have to be here in this PR) should be based on elementwise comparisons, and `[1, 2, 3].cycled(times: 6)` should be equivalent to `[1, 2, 3, 1, 2, 3].cycled(times: 3)`. Having `base` and `times` publicly accessible would raise the question of whether these properties are "salient" ones for the purposes of considering equivalence.

xwu (commented):
👍

Sources/Algorithms/Cycle.swift
LemonSpike: Should I also change `Cycle.base` to internal as well? This was why I used public here.

timvermeulen:
Thanks for taking this on @LemonSpike, this is a great start! Your `FiniteCycle` correctly delegates all of the work to an underlying `Product2<Range<Int>, Base>`. However, the real benefit will come from being able to efficiently implement all of the `Collection` methods that `Product2` already implements. By forwarding all public `Collection` and `BidirectionalCollection` methods to an underlying `Product2` collection, `FiniteCycle` will have all the same performance characteristics as `Product2`. Want to have a go at that?

LemonSpike:
Thanks for the feedback @timvermeulen! I have tried to add the `Collection` protocol conformances this time by forwarding to an underlying `Product2` collection. Are these what you had in mind? Thanks.

timvermeulen:
This is indeed what I had in mind, great work! The `Collection` conformance looks good, it implements all the relevant methods. There are just a few things left to do:
- We can get rid of the `Sequence` conformance entirely — normally it's needed in case the base type is only a `Sequence` and `Collection` cannot be implemented, but in the case of `FiniteCycle` we already require that `Base: Collection`, so it's no longer needed. In the same vein, the type needs `LazyCollectionProtocol` conformance rather than `LazySequenceProtocol`.
- `FiniteCycle` is still missing some conformances that `Product2` does have, implemented [here](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Product.swift#L433-L435).
- Most public methods are missing `@inlinable` annotations.
- The `Cycle.md` guide needs a slight tweak to reflect that `cycled(times:)` now has a new return type.

LemonSpike:
Hi @timvermeulen, thanks for the feedback. Re: your points, I tried to implement your suggestions as follows:

1. I removed the `Sequence` conformance entirely. To conform to `LazyCollectionProtocol` I had to conform to `LazySequenceProtocol` and `LazyCollectionProtocol` together where `Base: LazyCollectionProtocol`, because I couldn't inherit `LazySequenceProtocol` conformance automatically.
 2. Managed to add these as suggested, but because the `FiniteCycle.Index` is a typealias for `Product2.Index`, the `Index` `Hashable` conformance was already synthesized.
 3. Added `@inlinable` to all public methods, by making the `let product` `public`.
 4. Markdown file edited but reworded without `despite its name` since the name no longer includes `Sequence`.

Let me know your thoughts. Thanks.

timvermeulen:
Apologies for the delay, @LemonSpike!

> 1. I removed the `Sequence` conformance entirely. To conform to `LazyCollectionProtocol` I had to conform to `LazySequenceProtocol` and `LazyCollectionProtocol` together where `Base: LazyCollectionProtocol`, because I couldn't inherit `LazySequenceProtocol` conformance automatically.

We can get rid of the `Iterator` type and `makeIterator` method entirely. The typechecker did get confused about what the `Element` type is after I tried doing this, so you might need to add `typealias Element = Base.Element` to the `Collection` conformance.

> 2. Managed to add these as suggested, but because the `FiniteCycle.Index` is a typealias for `Product2.Index`, the `Index` `Hashable` conformance was already synthesized.

Great point! Let's for now give `FiniteCycle` its own `Index` type that wraps a `Product2.Index`, just so in the future we have the freedom to change the internals if the need ever arises.

> 3. Added `@inlinable` to all public methods, by making the `let product` `public`.

While we make all public and internal methods `@inlinable`, for stored properties `@usableFromInline` suffices, so there's no need to make the property public.

> 4. Markdown file edited but reworded without `despite its name` since the name no longer includes `Sequence`.

That looks good, could you also have it mention the conditional `LazyCollectionProtocol` conformance, similar to the `Cycle` description? And there's a note a bit higher up about `cycled(times:)` being a combination of two existing standard library functions which we can probably get rid of altogether.

> 5. I changed some `@usableFromInline` `internal` methods to `@inlinable` because of #109.

💯

Sources/Algorithms/Product.swift
```diff
@@ -44,7 +44,8 @@ extension Product2: Sequence {
     }
     
     @inlinable
-    public mutating func next() -> (Base1.Element, Base2.Element)? {
+    public mutating func next() -> (element1: Base1.Element,
+                                    element2: Base2.Element)? {
```
timvermeulen, line 48: Let's keep this tuple unlabeled — `Zip2Sequence`'s element tuple doesn't have labels either, and generic labels such as `element1` and `element2` don't really add any clarity over `.0` and `.1`.

LemonSpike: I am happy to do this, but I still think that the labels improve readability for users who don't know about tuple indexing. I saw we use more descriptive names for public APIs and type names, so I think it could help users understand the API. Let me know your thoughts.

timvermeulen: These labels don't improve the users' understanding, because they don't convey any useful information. It's impossible for us to give the tuple elements descriptive names because at this level of abstraction we have nothing to go on. To quote the [API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#naming):

>**Omit needless words.** Every word in a name should convey salient information at the use site.

Feel free to start a discussion about this topic on the forums, as this is equally relevant to the `Zip2Sequence` element type. For the time being we'd like to stick with the precedent set by the standard library.

natecook1000:
@swift-ci Please test

LemonSpike:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
Looks great, @LemonSpike — thanks for this addition! 🎉

PR #109, opened 2021-03-16
Inline all the things! 🎨

karwa:
#92 introduced quite a serious performance regression by refactoring inlinable code in to a helper function which was only _usable_ from inlined code, but not itself inlinable. The first commit corrects this.

Upon further inspection, I noticed a pattern - that _all_ internal functions seemed to only be `@usableFromInline` and not `@inlinable`, whereas their public siblings were all `@inlinable`. Inlining is critical for exactly this kind of code - lazy wrappers and collections and so on, and especially when they are generic. For instance, how is the client's instance of the compiler supposed to be able to optimise `Chain`, when the implementation of `offsetForward` is hidden from it?

So the second commit inlines basically everything that was previously `@usableFromInline`.

Additionally, I would strongly recommend setting up a benchmark suite, so that small refactorings don't cause huge performance regressions in future. I have some concerns about adding `swift-algorithms` as a dependency for my performance-sensitive project, and a benchmark suite which ensures at least stable baseline performance would help convince me that it's safe to rely on this project.

### DIFF
--- a/Sources/Algorithms/Chain.swift
+++ b/Sources/Algorithms/Chain.swift
@@ -19,7 +19,7 @@ public struct Chain2<Base1: Sequence, Base2: Sequence>
   /// The second sequence in this chain.
   public let base2: Base2
 
-  @usableFromInline
+  @inlinable
   internal init(base1: Base1, base2: Base2) {
     self.base1 = base1
     self.base2 = base2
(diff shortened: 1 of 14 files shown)


kylemacomber:
@swift-ci please test

natecook1000:
Thanks for chasing these down, @karwa! Agreed that this is a better default, and that we do eventually want a wait to keep benchmarks for this package.

PR #112, opened 2021-03-24
[Compacted] Adding `compacted()` method to remove all nils in a sequence or collection

### LINKED ISSUE #107: Add `compacted()` as a convenience for `lazy.compactMap { $0 }`
One of the most common uses of `compactMap` is to just flatten the `nil`s out of a sequence without transforming its elements, i.e. `x.lazy.compactMap { $0 }`.

A `compacted()` convenience method for this use case would be useful because, similar to `joined()` in the Standard Library, it wouldn't have to capture an escaping closure and (by the current conventions) it would be able to be lazy by default, i.e. `x.compacted() // => Compacted<[Int?], Int>`

A new `Compacted` sequence should also be able to conditionally conform to `Collection` and `BidirectionalCollection`, and it should [precompute on initialization](https://forums.swift.org/t/pitch-efficient-collection-startindex-throughout-the-standard-library/45317) in order to provide an O(1) `startIndex`.

Here's a sketch of the `Sequence` conformance:

```swift
struct Compacted<Base: Sequence, Element>: Sequence
  where Base.Element == Element?
{
  let base: Base

  struct Iterator: IteratorProtocol {
    var base: Base.Iterator

    mutating func next() -> Element? {
      while let wrapped = base.next() {
        if let some = wrapped {
          return some
        } else {
          // skip nil
        }
      }
      return nil
    }
  }

  func makeIterator() -> Iterator {
    return Iterator(base: base.makeIterator())
  }
}

extension Sequence {
  func compacted<Unwrapped>() -> Compacted<Self, Unwrapped> where Element == Unwrapped? {
    Compacted(base: self)
  }
}

var tests: [[Int?]] = [
  [],
  [0],
  [nil],
  [0, nil],
  [nil, 0],
  [0, nil, 1, nil, 2, nil],
  [0, 1, 2, nil, nil, nil],
  [nil, nil, nil, 0, 1, 2],
]

for test in tests {
  assert(test.compacted().elementsEqual(test.compactMap({ $0 })))
}

// let x: Array<Int> = [1, 2, 3]
// _ = x.compacted() // error: instance method 'compacted()' requires the types 'Int' and 'Unwrapped?' be equivalent
```

Thanks to @natecook1000 for coming up with how to make the types work out!

LucianoPAlmeida:
Fixes #107 

From the sketch from @kylemacomber on the issue, this just complement the implementation adding the tests and docs.
Let me know what you think =]

cc @kylemacomber @natecook1000 @timvermeulen 

Replace this paragraph with a description of your changes and rationale. Provide links to an existing issue or external references/discussions, if appropriate.

### Checklist
- [ ] I've read the [Contribution Guidelines](../CONTRIBUTING.md)

### DIFF
--- a/Sources/Algorithms/Compacted.swift
+++ b/Sources/Algorithms/Compacted.swift
@@ -197,6 +197,7 @@ extension CompactedSequence: Equatable
 
 extension CompactedSequence: Hashable
   where Element: Hashable {
+
   @inlinable
   public func hash(into hasher: inout Hasher) {
     for element in self {
--- a/Sources/Algorithms/Compacted.swift
+++ b/Sources/Algorithms/Compacted.swift
@@ -107,7 +107,7 @@ public struct CompactedCollection<Base: Collection, Element>: Collection
   
   @inlinable
   public func index(after i: Index) -> Index {
-    precondition(i < endIndex, "Index out of bounds")
+    precondition(i != endIndex, "Index out of bounds")
     
     let baseIdx = base.index(after: i.base)
     guard let idx = base[baseIdx...].firstIndex(where: { $0 != nil })
@@ -121,7 +121,7 @@ extension CompactedCollection: BidirectionalCollection
 
   @inlinable
   public func index(before i: Index) -> Index {
-    precondition(i > startIndex, "Index out of bounds")
+    precondition(i != startIndex, "Index out of bounds")
     
     guard let idx =
             base[startIndex.base..<i.base]
@@ -175,8 +175,6 @@ extension Collection {
 extension CompactedSequence: LazySequenceProtocol
   where Base: LazySequenceProtocol {}
 
-extension CompactedCollection: RandomAccessCollection
-  where Base: RandomAccessCollection {}
 extension CompactedCollection: LazySequenceProtocol
   where Base: LazySequenceProtocol {}
 extension CompactedCollection: LazyCollectionProtocol
--- a/Sources/Algorithms/Compacted.swift
+++ b/Sources/Algorithms/Compacted.swift
@@ -1,0 +1,227 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+/// A `Sequence` that iterates over every non-nil element from the original `Sequence`.
+public struct CompactedSequence<Base: Sequence, Element>: Sequence
+  where Base.Element == Element? {
+
+  @usableFromInline
+  let base: Base
+  
+  @inlinable
+  init(base: Base) {
+    self.base = base
+  }
+
+  public struct Iterator: IteratorProtocol {
+    @usableFromInline
+    var base: Base.Iterator
+    
+    @inlinable
+    init(base: Base.Iterator) {
+      self.base = base
+    }
+    
+    @inlinable
+    public mutating func next() -> Element? {
+      while let wrapped = base.next() {
+        guard let some = wrapped else { continue }
+        return some
+      }
+      return nil
+    }
+  }
+
+  @inlinable
+  public func makeIterator() -> Iterator {
+    return Iterator(base: base.makeIterator())
+  }
+}
+
+extension Sequence {
+  /// Returns a new `Sequence` that iterates over every non-nil element
+  /// from the original `Sequence`.
+  /// It produces the same result as `c.compactMap { $0 }`.
+  ///
+  ///     let c = [1, nil, 2, 3, nil]
+  ///     for num in c.compacted() {
+  ///         print(num)
+  ///     }
+  ///     // 1
+  ///     // 2
+  ///     // 3
+  ///
+  /// - Returns: A `Sequence` where the element is the unwrapped original
+  ///   element and iterates over every non-nil element from the original
+  ///   `Sequence`.
+  ///
+  /// Complexity: O(1)
+  @inlinable
+  public func compacted<Unwrapped>() -> CompactedSequence<Self, Unwrapped>
+    where Element == Unwrapped? {
+    CompactedSequence(base: self)
+  }
+}
+
+/// A `Collection` that iterates over every non-nil element from the original `Collection`.
+public struct CompactedCollection<Base: Collection, Element>: Collection
+  where Base.Element == Element? {
+
+  @usableFromInline
+  let base: Base
+  
+  @inlinable
+  init(base: Base) {
+    self.base = base
+    let idx = base.firstIndex(where: { $0 != nil }) ?? base.endIndex
+    self.startIndex = Index(base: idx)
+  }
+  
+  public struct Index {
+    @usableFromInline
+    let base: Base.Index
+    
+    @inlinable
+    init(base: Base.Index) {
+      self.base = base
+    }
+  }
+  
+  public var startIndex: Index
+  
+  @inlinable
+  public var endIndex: Index { Index(base: base.endIndex) }
+  
+  @inlinable
+  public subscript(position: Index) -> Element {
+    base[position.base]!
+  }
+  
+  @inlinable
+  public func index(after i: Index) -> Index {
+    precondition(i < endIndex, "Index out of bounds")
+    
+    let baseIdx = base.index(after: i.base)
+    guard let idx = base[baseIdx...].firstIndex(where: { $0 != nil })
+      else { return endIndex }
+    return Index(base: idx)
+  }
+}
+
+extension CompactedCollection: BidirectionalCollection
+  where Base: BidirectionalCollection {
+
+  @inlinable
+  public func index(before i: Index) -> Index {
+    precondition(i > startIndex, "Index out of bounds")
+    
+    guard let idx =
+            base[startIndex.base..<i.base]
+                .lastIndex(where: { $0 != nil })
+      else { fatalError("Index out of bounds") }
+    return Index(base: idx)
+  }
+}
+
+extension CompactedCollection.Index: Comparable {  
+  @inlinable
+  public static func < (lhs: CompactedCollection.Index,
+                        rhs: CompactedCollection.Index) -> Bool {
+    lhs.base < rhs.base
+  }
+}
+
+extension CompactedCollection.Index: Hashable
+  where Base.Index: Hashable {}
+
+extension Collection {
+  /// Returns a new `Collection` that iterates over every non-nil element
+  /// from the original `Collection`.
+  /// It produces the same result as `c.compactMap { $0 }`.
+  ///
+  ///     let c = [1, nil, 2, 3, nil]
+  ///     for num in c.compacted() {
+  ///         print(num)
+  ///     }
+  ///     // 1
+  ///     // 2
+  ///     // 3
+  ///
+  /// - Returns: A `Collection` where the element is the unwrapped original
+  ///   element and iterates over every non-nil element from the original
+  ///   `Collection`.
+  ///
+  /// Complexity: O(*n*) where *n*  is the number of elements in the
+  /// original `Collection`.
+  @inlinable
+  public func compacted<Unwrapped>() -> CompactedCollection<Self, Unwrapped>
+    where Element == Unwrapped? {
+    CompactedCollection(base: self)
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Protocol Conformances
+//===----------------------------------------------------------------------===//
+
+extension CompactedSequence: LazySequenceProtocol
+  where Base: LazySequenceProtocol {}
+
+extension CompactedCollection: RandomAccessCollection
+  where Base: RandomAccessCollection {}
+extension CompactedCollection: LazySequenceProtocol
+  where Base: LazySequenceProtocol {}
+extension CompactedCollection: LazyCollectionProtocol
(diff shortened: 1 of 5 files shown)


natecook1000 (commented):
Thanks for taking this one on, @LucianoPAlmeida! 👏 A few notes below about indexes and conformances.

Tests/SwiftAlgorithmsTests/CompactedTests.swift
```diff
@@ -0,0 +1,54 @@
...
+
+  var tests: [[Int?]] = [
+    [],
+    [0],
+    [nil],
+    [0, nil],
+    [nil, 0],
+    [0, nil, 1, nil, 2, nil],
+    [0, 1, 2, nil, nil, nil],
+    [nil, nil, nil, 0, 1, 2],
```
kylemacomber, line 25: Can we use `[nil, nil, nil, 0, 1, 2].uniquePermutations(ofCount: 0...)` here 😄?

LucianoPAlmeida:
@natecook1000 @kylemacomber Thank you for the review :)

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,174 @@
...
+    
+    var idx = base.index(after: i.base)
+    while idx != base.endIndex {
+      guard base[idx] != nil else {
+        base.formIndex(after: &idx)
+        continue
+      }
+      return Index(base: idx)
+    }
+    return Index(base: base.endIndex)
```
timvermeulen, line 108: You could avoid a raw loop here by using `firstIndex(while:)` like you already do with `_startIdx`, but called on a slice instead (starting at `base.index(after: i.base)`). And similarly in `index(before:)`, but with `lastIndex(where:)`.

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,174 @@
...
+        base.formIndex(before: &idx)
+        continue
+      }
+      return Index(base: idx)
+    }
+    return Index(base: base.startIndex)
+  }
+}
+
+extension CompactedCollection.Index: Comparable {
```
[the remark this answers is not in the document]
LucianoPAlmeida: Done!

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,174 @@
...
+}
+
+public struct CompactedCollection<Base: Collection, Element>: Collection
+  where Base.Element == Element? {
+
+  @usableFromInline
+  let base: Base
+  
+  @usableFromInline
+  let _startIdx: Base.Index
```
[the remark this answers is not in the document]
LucianoPAlmeida: Fixed! :)

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,174 @@
...
+  @inlinable
+  public subscript(position: Index) -> Element {
+    base[position.base]!
+  }
+  
+  @inlinable
+  public func index(after i: Index) -> Index {
+    guard i.base != base.endIndex else {
+      return Index(base: base.endIndex)
+    }
```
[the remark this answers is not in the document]
LucianoPAlmeida: Fixed! :)

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,174 @@
...
+}
+
+extension CompactedCollection: BidirectionalCollection
+  where Base: BidirectionalCollection {
+
+  @inlinable
+  public func index(before i: Index) -> Index {
+    guard i.base != base.startIndex else {
+      return Index(base: base.startIndex)
+    }
```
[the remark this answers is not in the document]
LucianoPAlmeida: Fixed! :)

Sources/Algorithms/Compacted.swift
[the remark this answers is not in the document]
LucianoPAlmeida: Done!

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,174 @@
...
+extension CompactedSequence: Equatable where Base: Equatable {}
+extension CompactedSequence: Hashable where Base: Hashable {}
+
+extension CompactedCollection: RandomAccessCollection
+  where Base: RandomAccessCollection {}
+extension CompactedCollection: LazySequenceProtocol
+  where Base: LazySequenceProtocol {}
+extension CompactedCollection: LazyCollectionProtocol
+  where Base: LazyCollectionProtocol {}
+extension CompactedCollection: Equatable where Base: Equatable {}
```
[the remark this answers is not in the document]
LucianoPAlmeida: Done! :)

Tests/SwiftAlgorithmsTests/CompactedTests.swift
LucianoPAlmeida: Fixed! :)

Sources/Algorithms/Compacted.swift
LucianoPAlmeida: I've adjusted, but actually I thought about it but just choose to use raw loop... but no strong feelings about it. Just fixed =]

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,195 @@
...
+
+  @inlinable
+  public func index(before i: Index) -> Index {
+    guard i != startIndex else {
+      fatalError("Index out of bounds")
+    }
+    
+    let baseIdx = base.index(before: i.base)
+    guard let idx = base[...baseIdx].lastIndex(where: { $0 != nil })
+      else { return Index(base: base.startIndex) }
```
natecook1000, line 113: This is actually an error condition, since the only index with _only_ `nil`s before it is `startIndex`, and the check at the beginning ensures we’re aren’t there.

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,195 @@
...
+  where Base: BidirectionalCollection {
+
+  @inlinable
+  public func index(before i: Index) -> Index {
+    guard i != startIndex else {
+      fatalError("Index out of bounds")
+    }
+    
+    let baseIdx = base.index(before: i.base)
+    guard let idx = base[...baseIdx].lastIndex(where: { $0 != nil })
```
natecook1000, line 112: We can just go from `i`'s position here:
```suggestion
   guard let idx = base[..<i.base].lastIndex(where: { $0 != nil })
```

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,195 @@
...
+// Hashable and Equatable conformance are based on each non-nil
+// element on base collection.
+extension CompactedSequence: Equatable
+  where Base: Equatable, Base.Element: Equatable {
+
+  @inlinable
+  public static func ==(lhs: CompactedSequence,
+                        rhs: CompactedSequence) -> Bool {
+    lhs.compactMap({ $0 })
+      .elementsEqual(rhs.compactMap({ $0 }))
```
natecook1000, line 159: I don't think you need the `compactMap` here — did you get an error with that? (Same for the collection version)

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+
+extension CompactedSequence: Hashable
+  where Base: Hashable, Base.Element: Hashable {
+  @inlinable
+  public func hash(into hasher: inout Hasher) {
+    for element in base {
```
natecook1000, line 167: This could just be `for element in self`, so that you don't need to re-unwrap. (Ditto with the collection version)

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,195 @@
...
+extension CompactedCollection: LazySequenceProtocol
+  where Base: LazySequenceProtocol {}
+extension CompactedCollection: LazyCollectionProtocol
+  where Base: LazyCollectionProtocol {}
+
+
+// Hashable and Equatable conformance are based on each non-nil
+// element on base collection.
+extension CompactedSequence: Equatable
+  where Base: Equatable, Base.Element: Equatable {
```
natecook1000, line 153: Since we're implementing these conformances directly, we shouldn't need to have the `Base` conformance in these `where` clauses. (Here and in the other conditional conformances.)

Tests/SwiftAlgorithmsTests/CompactedTests.swift
```diff
@@ -0,0 +1,68 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
```
natecook1000, line 5: Nit:
```suggestion
// Copyright (c) 2021 Apple Inc. and the Swift project authors
```

natecook1000 (commented):
Looking good, @LucianoPAlmeida! Some more notes for you below, then I think we just need docs on the methods themselves.

Sources/Algorithms/Compacted.swift
LucianoPAlmeida: Sorry, I was thinking in terms of `lhs.base` and `rhs.base`... 😓

LucianoPAlmeida:
@natecook1000 Thanks again and sorry for those mistakes, I think all things were addressed =]

LucianoPAlmeida:
@swift-ci Please test

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,228 @@
...
+
+//===----------------------------------------------------------------------===//
+// Protocol Conformances
+//===----------------------------------------------------------------------===//
+
+extension CompactedSequence: LazySequenceProtocol
+  where Base: LazySequenceProtocol {}
+
+extension CompactedCollection: RandomAccessCollection
+  where Base: RandomAccessCollection {}
```
natecook1000, line 179: I missed this — because we don't know how far apart the non-`nil` elements are until we iterate, we can't make `CompactedCollection` conform to `RandomAccessCollection`.

Tests/SwiftAlgorithmsTests/CompactedTests.swift
```diff
@@ -0,0 +1,68 @@
...
+  func testCompactedBidirectionalCollection() {
+    for array in self.tests {
+      XCTAssertEqualSequences(array.compactMap({ $0 }).reversed(),
+                              array.compacted().reversed())
+    }
+  }
+  
+  func testCollectionTraversals() {
+    for array in self.tests {
+      validateIndexTraversals(array)
```
natecook1000, line 41: Looks like we need to call `compacted()` here.

Tests/SwiftAlgorithmsTests/CompactedTests.swift
```diff
@@ -0,0 +1,68 @@
...
+    for array in self.tests {
+      let seq = array.eraseToAnyHashableSequence()
+      XCTAssertEqualHashValue(
+        seq.compacted(), seq.compactMap({ $0 }).compacted()
+      )
+      XCTAssertEqualHashValue(
+        array.compacted(), array.compactMap({ $0 }).compacted()
+      )
+    }
+  }
```
natecook1000, line 67: For these two tests it would be great to make sure that equality and hashing work for non-equal base sequences/collections that are equal when compacted. E.g. `[1, 2, 3, nil, nil, 4].compacted() == [1, nil, 2, nil, 3, 4].compacted()`. You're covering the correctness of compacting already, so I don't think you need to do the comparison with `compactMap { $0 }` any more.

natecook1000 (commented):
Almost there! Just a couple notes left.

Sources/Algorithms/Compacted.swift
LucianoPAlmeida: Makes sense :)

Tests/SwiftAlgorithmsTests/CompactedTests.swift
LucianoPAlmeida: Sorry for that

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,226 @@
...
+  public var endIndex: Index { Index(base: base.endIndex) }
+  
+  @inlinable
+  public subscript(position: Index) -> Element {
+    base[position.base]!
+  }
+  
+  @inlinable
+  public func index(after i: Index) -> Index {
+    precondition(i < endIndex, "Index out of bounds")
```
timvermeulen, line 110:
```suggestion
    precondition(i != endIndex, "Index out of bounds")
```

We try to compare indices using `!=` wherever possible to be consistent with the stdlib, also see https://github.com/apple/swift-algorithms/pull/65#issuecomment-766100210

Sources/Algorithms/Compacted.swift
```diff
@@ -0,0 +1,226 @@
...
+    return Index(base: idx)
+  }
+}
+
+extension CompactedCollection: BidirectionalCollection
+  where Base: BidirectionalCollection {
+
+  @inlinable
+  public func index(before i: Index) -> Index {
+    precondition(i > startIndex, "Index out of bounds")
```
timvermeulen, line 124:
```suggestion
    precondition(i != startIndex, "Index out of bounds")
```

Same as above 🙂

Sources/Algorithms/Compacted.swift
LucianoPAlmeida: Ah ok, thanks

LucianoPAlmeida: Fixed! :)

natecook1000:
@swift-ci Please test

LucianoPAlmeida:
@swift-ci Please test

natecook1000:
:shipit:

PR #114, opened 2021-03-25
Inline all the reductions things!

danielctull:
As mentioned by @karwa in #109, functions should be marked as `@inlinable`. Due to timing of the two pull requests, the `reductions` addition hadn't taken this change into account. This change resolves that by making all the internal functions `@inlinable`.


timvermeulen:
@swift-ci please test

danielctull:
Thanks @timvermeulen! 🙂

PR #117, opened 2021-03-26
Use `uniquePermutations(ofCount:)` in lazy split test.

toddthomas:
Concise! Eliminates the big, unwieldy array literals.

### Expanded test coverage...
Currently testing all subsets of length 0 through 10 of the sequence `||||||||||EEEEEEEEEE` (`|`: separator; `E`: non-separator). See the comment above `testAllLength0Through10()` for the reasoning. That's quite a bit more coverage than the previous revision, which exhaustively tested only sequences of length 0 through 4, plus the unique permutations of select sequences of lengths 5 through 9.

### ...at a cost
Testing 2047 unique sequences has a cost of 3x longer test runtime. The lazy split tests alone now take almost 8 seconds on my machine. Total swift-algorithms test runtime is around 12 seconds, compared to 4 previously. If that's unacceptable, reducing the coverage to all possible patterns of lengths 0-9 reduces the lazy split test runtime to less than 2 seconds, and is still more extensive than the previous revision.

I can't make an argument that it's essential to cover all sequences up through length 10. I like that length 10 allows for certain interesting test cases like `||EE||EE||` (multiple adjacent separators at beginning, middle and end, sandwiching subsequences of multiple elements), and `E||EE||E||` (adding an element to the beginning of a pattern quite similar to the previous one).

It's more that `uniquePermutations(ofCount:)` makes it so easy to add that much coverage, one might say, "why not?" To which another might reply, "because it takes too long to run." And I'm fine with either argument.

### Enhanced validation
Because we're now passing a great variety of different patterns to the same invocation of `Validator.validate()`, `Validator` has been modified to compute an interesting `maxSplits` for each tested case if none is provided at initialization.

### Checklist

### DIFF
--- a/Tests/SwiftAlgorithmsTests/LazySplitTests.swift
+++ b/Tests/SwiftAlgorithmsTests/LazySplitTests.swift
@@ -17,403 +17,41 @@ import XCTest
 //===----------------------------------------------------------------------===//
 
 final class LazySplitTests: XCTestCase {
-  func testEmpty() {
-    Validator(
-      subject: [],
-      separator: .element(0),
-      maxSplits: 1
-    ).validate()
-  }
-
-  // The following test names indicate the sequence being split, using the
-  // notation
-  // S: Separator
-  // E: Element
+  // Sequences to be split are patterns of separators and non-separators. Non-
+  // separators are hereafter referred to as "elements". All elements are
+  // equivalent from the algorithm's perspective, so the patterns can be
+  // represented as strings of "E" for elements, and "|" for separators. For
+  // example, ||EE is a pattern that begins with two separators and ends with
+  // two elements.
   //
-  // Values of `maxSplit` are interesting when they're less than the number of
-  // splits that would occur with the default. That varies depending on whether
-  // empty subsequences are omitted or included. `Validator` tests both the
-  // default and the value provided, but there are cases below where the
-  // provided `maxSplit` isn't less than the number of splits when empty
-  // subsequences are omitted.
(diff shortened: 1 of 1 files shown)


natecook1000:
@swift-ci Please test

kylemacomber:
I think we should cut down on the time these tests take. The tests have already verified (to the extent they can) the correctness of the existing implementation. Going forward the tests will serve to catch regressions. But I think we will rarely change the implementation of lazy split, so 12 seconds (and even 2 seconds!) seems egregious.

If there are specific interesting cases, like the ones of length 10 you call out, maybe we should hardcode those in addition to an exhaustive coverage of shorter sequences, say of length 0-6?

toddthomas:
> I think we should cut down on the time these tests take.

I'll investigate doing something like what you suggest. Thanks for taking a look!

Tests/SwiftAlgorithmsTests/LazySplitTests.swift
```diff
@@ -560,13 +206,20 @@ final class LazySplitTests: XCTestCase {
...
     }
 
-    private func _validateAsSequence<T: Sequence>(_ s: T)
+    private func validateAsSequence<T: Sequence>(_ s: T)
     where T.Element == C.Element {
       // Default max splits, omitting empty sequences
+      var testSplitCountOmittingEmpties: Int
       switch separator {
       case let .element(element):
         let expected = s.split(separator: element).map { Array($0) }
```
kocsma, line 215: >as

PR #119, opened 2021-03-28
Introduce `adjacentPairs`

mpangburn:
Implements an `adjacentPairs` algorithm on Sequence, per forum discussion: https://forums.swift.org/t/add-an-adjacentpairs-algorithm-to-sequence/14817.

### Description
Lazily iterates over tuples of adjacent elements.

This operation is available for any sequence by calling the `adjacentPairs()` method.

```swift
let numbers = (1...5)
let pairs = numbers.adjacentPairs()
// Array(pairs) == [(1, 2), (2, 3), (3, 4), (4, 5)]
```

### Detailed Design
The `adjacentPairs()` method is declared as a `Sequence` extension returning `AdjacentPairs`.

```swift
extension Sequence {
  public func adjacentPairs() -> AdjacentPairs<Self>
}
```

The resulting `AdjacentPairs` type is a sequence, with conditional conformance to `Collection`, `BidirectionalCollection`, and `RandomAccessCollection` when the underlying sequence conforms.

The spelling `zip(s, s.dropFirst())` for a sequence `s` is an equivalent operation on collection types; however, this implementation is undefined behavior on single-pass sequences, and `Zip2Sequence` does not conditionally conform to the `Collection` family of protocols.

### Documentation Plan
- API introduced in AdjacentPairs.swift is documented.
- AdjacentPairs.md is included in Guides.
- README is updated.
- CHANGELOG is updated.

### Test Plan
Unit tests capture expected index-related invariants about `Collection` and verify expected behavior for sequences of varying lengths, including the empty and single-element sequence cases.

### Source Impact
This change is strictly additive.

### Checklist

### DIFF
--- a/Sources/Algorithms/AdjacentPairs.swift
+++ b/Sources/Algorithms/AdjacentPairs.swift
@@ -12,7 +12,7 @@
 extension Sequence {
   /// Creates a sequence of adjacent pairs of elements from this sequence.
   ///
-  /// In the `AdjacentPairs` instance returned by this method, the elements of
+  /// In the `AdjacentPairsSequence` returned by this method, the elements of
   /// the *i*th pair are the *i*th and *(i+1)*th elements of the underlying
   /// sequence.
   /// The following example uses the `adjacentPairs()` method to iterate over
@@ -25,14 +25,37 @@ extension Sequence {
   ///    // Prints "(2, 3)"
   ///    // Prints "(3, 4)"
   ///    // Prints "(4, 5)"
-  public func adjacentPairs() -> AdjacentPairs<Self> {
-    AdjacentPairs(_base: self)
+  @inlinable
+  public func adjacentPairs() -> AdjacentPairsSequence<Self> {
+    AdjacentPairsSequence(base: self)
+  }
+}
+
+extension Collection {
+  /// A collection of adjacent pairs of elements built from an underlying collection.
+  ///
+  /// In an `AdjacentPairsCollection`, the elements of the *i*th pair are the *i*th
+  /// and *(i+1)*th elements of the underlying sequence. The following example
+  /// uses the `adjacentPairs()` method to iterate over adjacent pairs of
+  /// integers:
+  /// ```
+  /// for pair in (1...5).adjacentPairs() {
+  ///     print(pair)
+  /// }
+  /// // Prints "(1, 2)"
+  /// // Prints "(2, 3)"
+  /// // Prints "(3, 4)"
+  /// // Prints "(4, 5)"
+  /// ```
+  @inlinable
+  public func adjacentPairs() -> AdjacentPairsCollection<Self> {
+    AdjacentPairsCollection(base: self)
   }
 }
 
 /// A sequence of adjacent pairs of elements built from an underlying sequence.
 ///
-/// In an `AdjacentPairs` sequence, the elements of the *i*th pair are the *i*th
+/// In an `AdjacentPairsSequence`, the elements of the *i*th pair are the *i*th
 /// and *(i+1)*th elements of the underlying sequence. The following example
 /// uses the `adjacentPairs()` method to iterate over adjacent pairs of
 /// integers:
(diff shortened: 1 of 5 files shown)


ollieatkinson:
We have something that can do this already:

```swift
import Algorithms

let numbers = (1...5)
let pairs = numbers.windows(ofCount: 2)
// [[1, 2], [2, 3], [3, 4], [4, 5]]
```

If you wanted to get your exact API for tuples you can write an extension re-using the windows(ofCount:) implementation: 
```swift
extension Collection {
    
    func adjacentPairs() -> AnyCollection<(Element, Element)> {
        AnyCollection(windows(ofCount: 2).map {
            ($0[$0.startIndex], $0[$0.index(after: $0.startIndex)])
        })
    }
}
```

natecook1000:
@ollieatkinson While there's overlap between this and `windows(ofCount:)`, `adjacentPairs` provides enough benefit that it can stand as its own, separate adapter. Primarily, the fact that it vends tuples as its element type simplifies a variety of pairwise operations — for example, you can write test that a sequence is sorted by writing `seq.adjacentPairs().allSatisfy(<=)`.

natecook1000:
One of the things we're finding as we build the wrappers in this package is that a collection's `startIndex`, `endIndex`, and the subscript really do need to be O(1), as documented (even though we haven't always adhered to that in the stdlib itself). When that isn't the case, combined operations can end up with unpredictable performance.

For this collection wrapper, that means that we'll need to compute `startIndex` when calling `adjacentPairs()` and store it as an instance property, instead of calculating the first pair of indices each time. _That_, in turn, will mean we need separate `AdjacentPairsSequence` and `AdjacentPairsCollection` types, and that the index for the collection type should wrap both indices instead of just the first one. With those changes, you'll be able to access the start/end index and use the subscript without needing to advance indices on the underlying collection.

ollieatkinson:
> One of the things we're finding as we build the wrappers in this package is that a collection's startIndex, endIndex, and the subscript really do need to be O(1), as documented (even though we haven't always adhered to that in the stdlib itself). When that isn't the case, combined operations can end up with unpredictable performance.

This is a good call, and something which is worthy to be wary of - when implementing the sliding window algorithm it was one of the primary concerns that access to both `startIndex` and `endIndex` is O(1).

It's true that the current implementation of the sliding window algorithm stores the starting offset, I would be interested to know if the impact of that storage is causing trouble for many algorithms given the fantastic capacity of our devices today. 

I am a little unsure to when we stop writing new wrappers vs using composition of existing types, and improving those to unlock new capabilities - for example, improving the sliding windows algorithms to address any of the concerns we would have to re-use in a implementation of `adjacentPairs`.

Having said all of this, for what it's worth, the implementation of `adjacentPairs` here looks good!

timvermeulen:
>It's true that the current implementation of the sliding window algorithm stores the starting offset, I would be interested to know if the impact of that storage is causing trouble for many algorithms given the fantastic capacity of our devices today.

The end user likely won't notice the difference in most scenarios, but in extreme cases it can make a meaningful difference (as described [here](https://forums.swift.org/t/pitch-efficient-collection-startindex-throughout-the-standard-library/45317)).

>I am a little unsure to when we stop writing new wrappers vs using composition of existing types, and improving those to unlock new capabilities - for example, improving the sliding windows algorithms to address any of the concerns we would have to re-use in a implementation of `adjacentPairs`.

We should certainly compose existing abstractions to make new ones whenever we reasonably can! And `slidingWindows` indeed almost gets us there. Unfortunately `Sequence` and `Collection` are sufficiently different from each other that in this case we're unable to create a `Sequence` method on top of one that's only available to `Collection`.

mpangburn:
@natecook1000 Appreciate the pointers—I've made the following changes:
- `AdjacentPairsSequence` and `AdjacentPairsCollection` are distinct types, and the `adjacentPairs()` extension method is overloaded appropriately. 
- For `AdjacentPairsCollection`, `startIndex` is precomputed, and `AdjacentPairsCollection.Index` stores both indices into the underlying collection.
- `AdjacentPairsSequence.Iterator` no longer computes `base.next()` in `init`, deferring fetching the initial element until its own `next` method is called.
- Inlining annotations have been added.

Let me know if there are any other improvements I can make.

Sources/Algorithms/AdjacentPairs.swift
```diff
@@ -0,0 +1,271 @@
...
+extension AdjacentPairsCollection: Collection {
+  @inlinable
+  public var endIndex: Index {
+    switch base.endIndex {
+    case startIndex.first, startIndex.second:
+      return startIndex
+    case let end:
+      return Index(first: end, second: end)
+    }
+  }
```
timvermeulen, line 200: I strongly suggest unconditionally representing `endIndex` as `Index(first: base.endIndex, second: base.endIndex)`, and adapting `startIndex` to match this representation in the edge case that `base.count == 1` (instead of the other way around). This is the approach we take in `Windows` as well. Having a consistent representation of `endIndex` often makes it easier to reason about index manipulation logic, and I think you'll find that it will improve some of your code.

Sources/Algorithms/AdjacentPairs.swift
```diff
@@ -0,0 +1,271 @@
...
+      let secondOffsetIndex = base.index(after: firstOffsetIndex)
+      return secondOffsetIndex == base.endIndex
+        ? endIndex
+        : Index(first: firstOffsetIndex, second: secondOffsetIndex)
+    } else {
+      return i == endIndex
+        ? Index(first: base.index(i.first, offsetBy: distance - 1),
+                second: base.index(i.first, offsetBy: distance))
+        : Index(first: base.index(i.first, offsetBy: distance),
+                second: i.first)
```
timvermeulen, line 230: You could avoid computing the first and second indices separately here, since you can cheaply compute the firts if you already have the second.

Sources/Algorithms/AdjacentPairs.swift
```diff
@@ -0,0 +1,271 @@
...
+    @inlinable
+    internal init(first: Base.Index, second: Base.Index) {
+      self.first = first
+      self.second = second
+    }
+
+    @inlinable
+    public static func < (lhs: Index, rhs: Index) -> Bool {
+      (lhs.first, lhs.second) < (rhs.first, rhs.second)
+    }
```
timvermeulen, line 187: You should only need to compare one of the two underlying indices here instead of both, and the same applies to `==`.

natecook1000:
@swift-ci Please test

natecook1000:
Going to merge now and follow up — connected with @mpangburn offline (thank you! :clap:)

PR #124, opened 2021-04-09
Remove collection conformances to `Equatable` and `Hashable`

timvermeulen:
Implementing `Equatable` (and especially `Hashable`) correctly isn't straightforward for some of the non-trivial collection wrappers in this package, if we consider collections to be equal if they produce the same elements. Here are some examples of comparisons that should evaluate to `true`, despite the collections having unequal base collections:

```swift
[1, 3, 5].striding(by: 1)         == [1, 2, 3, 4, 5].striding(by: 2)
[1, 2].windows(ofCount: 10)       == [1, 2, 3].windows(ofCount: 100)
product([1, 2], "")               == product([1, 2, 3], "")
chain([1, 2], 3..<6)              == chain([1, 2, 3], 4..<6)
[1, 2].combinations(ofCount: 0)   == [1, 2, 3].combinations(ofCount: 0)
[1, 1, 1, 2].uniquePermutations() == [1, 2].uniquePermutations()
[1, 1, 1].cycled(times: 2)        == [1, 1].cycled(times: 3)
```

Implementing `Hashable` is often even harder because equal values should produce equal hash values, and therefore every collection needs to hash in the exact same way as all other collections equal to it.

For the moment we're removing all conformances of the sequence and collection types to `Equatable` and `Hashable` to avoid any ambiguity on this front, which is consistent with sequence and collection adaptors in the standard library. Most of the time `s1.elementsEqual(s2)` should be preferred anyway, because it lets you compare sequences that aren't instances of the same type.

### Checklist


timvermeulen:
@swift-ci please test

natecook1000:
@swift-ci Please test

natecook1000 (approved):
> +2 −209 

🎉

PR #141, opened 2021-05-07
Adjacent pairs wrapping

mdznr:
### Description
Adds an optional `wrapping` parameter to `adjacentPairs` algorithm that allows you to include a pair of the last and first element.

### Detailed Design
```swift
extension Sequence {
  /// …
  ///
  /// The following example uses `adjacentPairs(wrapping: true)` to iterate over
  /// adjacent pairs of integers, including the last and first:
  ///
  ///     for pair in (1...).prefix(5).adjacentPairs(wrapping: true) {
  ///         print(pair)
  ///     }
  ///     // Prints "(1, 2)"
  ///     // Prints "(2, 3)"
  ///     // Prints "(3, 4)"
  ///     // Prints "(4, 5)"
  ///     // Prints "(5, 1)"
  ///
  /// - Parameter wrapping: If `true`, include the pair of the last and first
  /// elements of the sequence as the last elements of the returned collection.
  /// Defaults to `false`.
  @inlinable
  public func adjacentPairs(wrapping: Bool = false) -> AdjacentPairsSequence<Self> {
    AdjacentPairsSequence(base: self, wrapping: wrapping)
  }
}

extension Collection {
  /// …
  ///
  /// - Parameter wrapping: If `true`, include the pair of the last and first
  /// elements of the collection as the last elements of the returned
  /// collection. Defaults to `false`.
  @inlinable
  public func adjacentPairs(wrapping: Bool = false) -> AdjacentPairsCollection<Self> {
    AdjacentPairsCollection(base: self, wrapping: wrapping)
  }
}
```

### Alternatives Considered
- A new `Sequence` and `Collection` algorithm. This would have resulted in a lot of duplicate code from `AdjacentPairs` that would be better shared between the implementations.

### Documentation Plan
I’ll waiting to hear feedback on whether or not this is desirable and on the direction before diving into the documentation too much.
- [ ] Improve inline documentation
- [ ] Update `Guides/AdjacentPairs.md`
- [ ] `README.md`

### Test Plan
For each of the existing `adjacentPairs` unit tests, I’ve added one that tests the wrapping behavior as well.

### Source Impact
Existing calls to `adjacentPairs()` continues to work as-is since the `wrapping` parameter has a default value of `false`.

### Checklist

### DIFF
--- a/Sources/Algorithms/AdjacentPairs.swift
+++ b/Sources/Algorithms/AdjacentPairs.swift
@@ -144,7 +144,7 @@ public struct AdjacentPairsCollection<Base: Collection> {
       return
     }
     
-    // If there's only one element (i.e. the second index of base == endIndex)
+    // If there's only one element (i.e., the second index of base == endIndex),
     // then this collection should be empty.
     let secondIndex = base.index(after: base.startIndex)
     self.startIndex = secondIndex == base.endIndex
--- a/Sources/Algorithms/AdjacentPairs.swift
+++ b/Sources/Algorithms/AdjacentPairs.swift
@@ -64,9 +64,13 @@ extension Collection {
   ///     // Prints "(2, 3)"
   ///     // Prints "(3, 4)"
   ///     // Prints "(4, 5)"
+  ///
+  /// - Parameter wrapping: If `true`, include the pair of the last and first
+  /// elements of the collection as the last element of the returned
+  /// collection. Defaults to `false`.
   @inlinable
-  public func adjacentPairs() -> AdjacentPairsCollection<Self> {
-    AdjacentPairsCollection(base: self)
+  public func adjacentPairs(wrapping: Bool = false) -> AdjacentPairsCollection<Self> {
+    AdjacentPairsCollection(base: self, wrapping: wrapping)
   }
 }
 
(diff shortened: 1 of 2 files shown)


xwu:
`wrapping` suggests to me an infinite sequence (which I'd imagine has its own uses); is there potentially another way of describing this specific option?

kylemacomber:
I appreciate that this was motivated by a concrete real world use case, however after thinking about this more, I'm not sure adding a `wrapping` flag (regardless of its name) pulls its weight.

For example, I think this behavior can be relatively trivially composed out of `chain(x, x.prefix(1)).adjacentPairs()`:

```swift
let x = (1...).prefix(5)
for pair in chain(x, x.prefix(1)).adjacentPairs() {
    print(pair)
}
// Prints "(1, 2)"
// Prints "(2, 3)"
// Prints "(3, 4)"
// Prints "(4, 5)"
// Prints "(5, 1)"
```

I vote that we close this for now and revisit if additional concrete use cases arise.

natecook1000:
Agreed, we can close this one for now. Thanks for testing this out, @mdznr — happy to discuss this more on the forums!

PR #143, opened 2021-05-12
Minor revisions to `minAndMax()`

natecook1000:
This removes the `#if`s to land on a single implementation for `minAndMax()` and adds a couple of additional tests.

### Checklist

### DIFF
--- a/Sources/Algorithms/MinMax.swift
+++ b/Sources/Algorithms/MinMax.swift
@@ -440,55 +440,16 @@ extension Sequence {
     // Confirm the initial bounds.
     if try areInIncreasingOrder(highest, lowest) { swap(&lowest, &highest) }
 
-    #if true
     // Read the elements in pairwise.  Structuring the comparisons around this
     // is actually faster than loops based on extracting and testing elements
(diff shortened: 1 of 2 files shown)


natecook1000:
@swift-ci Please test

Sources/Algorithms/MinMax.swift
```diff
@@ -440,55 +440,16 @@ extension Sequence {
...
-    }
-
-    // Read the elements in four-at-a-time.  Some say this is more effective
-    // than a two-at-a-time loop.
-    while let a = iterator.next() {
-      let b = iterator.next() ?? a
-      let c = iterator.next() ?? b
-      let d = iterator.next() ?? c
-      let (low, high) = try minAndMaxOf4(a, b, c, d)
+      var high = iterator.next() ?? low
```
CTMacUser, line 447: Copying `low` as `high` ended up being faster than branching when the latter would be `nil`?

natecook1000: IIRC it didn't cost anything, and the following code is simpler.

PR #152, opened 2021-07-15
Add `partitioned(_:)`

mdznr:
### Description
Adds a `partitioned(_:)` algorithm. This is very similar to `filter(_:)`, but instead of just getting an array of the elements that _do_ match a given predicate, also get a second array for the elements that _did not_ match the same predicate.

This is more performant than calling `filter(_:)` twice on the same input with mutually-exclusive predicates since:
1. It only requires a single pass of the elements
2. If the input has a known number of elements, the cumulative space for both returned arrays is known and can avoid array buffer resizing.

```swift
let cast = ["Vivien", "Marlon", "Kim", "Karl"]
let (longNames , shortNames) = cast.partitioned({ $0.count < 5 })
print(longNames)
// Prints "["Vivien", "Marlon"]"
print(shortNames)
// Prints "["Kim", "Karl"]"
```

### Detailed Design
```swift
extension Sequence {
  @inlinable
  public func partitioned(
    _ belongsInSecondCollection: (Element) throws -> Bool
  ) rethrows -> ([Element], [Element])
}
```

### Naming
At a high-level, this acts similarly to the `partition(_:)` family of functions in that it separates all the elements in a given collection in two parts, those that do and do not match a given predicate. Thanks, @timvermeulen for help with naming!

### Documentation Plan
- Inline documentation for each new function
- Comments in the implementation
- Updated README.md
- Added to Guides/Partitioned.md

### Test Plan
- Adds unit tests for example given in documentation
- Adds unit tests for various inputs
- Adds unit tests for empty input

### Source Impact
This is purely additive

### Checklist

### DIFF
--- a/Guides/Partition.md
+++ b/Guides/Partition.md
@@ -42,6 +42,32 @@ let p = numbers.partitioningIndex(where: { $0.isMultiple(of: 20) })
 // numbers[p...] = [20, 40, 60]
 ```
 
+The standard library’s existing `filter(_:)` method provides functionality to
+get the elements that do match a given predicate. `partitioned(_:)` returns
+both the elements that match the preciate as well as those that don’t, as a
+tuple.
+
+```swift
+let cast = ["Vivien", "Marlon", "Kim", "Karl"]
+let (longNames , shortNames) = cast.bifurcate({ $0.count < 5 })
+print(longNames)
+// Prints "["Vivien", "Marlon"]"
+print(shortNames)
+// Prints "["Kim", "Karl"]"
+```
+
+There’s also a function to bifurcate a collection into a prefix and a suffix, up
+to but not including a given index:
+
+```swift
+let cast = ["Vivien", "Marlon", "Kim", "Karl"]
+let (callbacks, alternates) = cast.bifurcate(upTo: 2)
+print(callbacks)
+// Prints "["Vivien", "Marlon"]"
+print(alternates)
+// Prints "["Kim", "Karl"]"
+```
+
 ## Detailed Design
 
 All mutating methods are declared as extensions to `MutableCollection`.
@@ -69,11 +95,17 @@ extension Collection {
         where belongsInSecondPartition: (Element) throws -> Bool
     ) rethrows -> Index
 }
+
+extension Sequence {
+	public func bifurcate(
+	  _ belongsInFirstCollection: (Element) throws -> Bool
+	) rethrows -> ([Element], [Element])
+}
 ```
 
 ### Complexity
 
-The existing partition is an O(_n_) operations, where _n_ is the length of the
+The existing partition is an O(_n_) operation, where _n_ is the length of the
 range to be partitioned, while the stable partition is O(_n_ log _n_). Both
 partitions have algorithms with improved performance for bidirectional
 collections, so it would be ideal for those to be customization points were they
@@ -81,6 +113,9 @@ to eventually land in the standard library.
 
 `partitioningIndex(where:)` is a slight generalization of a binary search, and
 is an O(log _n_) operation for random-access collections; O(_n_) otherwise.
+
+`partitioned(_:)` is an O(_n_) operation, where _n_ is the number of elements in
+the original sequence.
 
 ### Comparison with other languages
 
--- a/Sources/Algorithms/Partition.swift
+++ b/Sources/Algorithms/Partition.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Algorithms open source project
 //
-// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -204,3 +204,156 @@ extension Collection {
   }
 }
 
+//===----------------------------------------------------------------------===//
+// partitioned(_:)
+//===----------------------------------------------------------------------===//
+
+extension Sequence {
+  /// Returns two arrays containing, in order, the elements of the sequence that
+  /// do and don’t satisfy the given predicate, respectively.
+  ///
+  /// In this example, `partitioned(_:)` is used to separate the input based on
+  /// names that aren’t and are shorter than five characters, respectively:
+  ///
+  ///     let cast = ["Vivien", "Marlon", "Kim", "Karl"]
+  ///     let (longNames, shortNames) = cast.partitioned({ $0.count < 5 })
+  ///     print(longNames)
+  ///     // Prints "["Vivien", "Marlon"]"
+  ///     print(shortNames)
+  ///     // Prints "["Kim", "Karl"]"
+  ///
+  /// - Parameter belongsInSecondCollection: A closure that takes an element of
+  /// the sequence as its argument and returns a Boolean value indicating
+  /// whether the element should be included in the second returned array.
+  /// Otherwise, the element will appear in the first returned array.
+  ///
+  /// - Returns: Two arrays with with all of the elements of the receiver. The
+  /// first array contains all the elements that `belongsInSecondCollection`
+  /// didn’t allow, and the second array contains all the elements that
+  /// `belongsInSecondCollection` allowed.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the sequence.
+  ///
+  /// - Note: This algorithm performs a bit slower than the same algorithm on
+  /// `RandomAccessCollection` since the size of the sequence is unknown, unlike
+  /// `RandomAccessCollection`.
+  @inlinable
+  public func partitioned(
+	_ belongsInSecondCollection: (Element) throws -> Bool
+  ) rethrows -> ([Element], [Element]) {
+	var lhs = ContiguousArray<Element>()
+	var rhs = ContiguousArray<Element>()
+	
+	for element in self {
+	  if try belongsInSecondCollection(element) {
+		rhs.append(element)
+	  } else {
+		lhs.append(element)
+	  }
+	}
+	
+	return _tupleMap((lhs, rhs), { Array($0) })
+  }
+}
+
+extension Collection {
+  // This is a specialized version of the same algorithm on `Sequence` that
+  // avoids reallocation of arrays since `count` is known ahead of time.
+  @inlinable
+  public func partitioned(
+	_ belongsInSecondCollection: (Element) throws -> Bool
+  ) rethrows -> ([Element], [Element]) {
+	guard !self.isEmpty else {
+	  return ([], [])
+	}
+	
+	// Since `RandomAccessCollection`s have known sizes (access to `count` is
+	// constant time, O(1)), we can allocate one array of size `self.count`,
+	// then insert items at the beginning or end of that contiguous block. This
+	// way, we don’t have to do any dynamic array resizing. Since we insert the
+	// right elements on the right side in reverse order, we need to reverse
+	// them back to the original order at the end.
+	
+	let count = self.count
+	
+	// Inside of the `initializer` closure, we set what the actual mid-point is.
+	// We will use this to partitioned the single array into two in constant time.
+	var midPoint: Int = 0
+	
+	let elements = try [Element](
+	  unsafeUninitializedCapacity: count,
+	  initializingWith: { buffer, initializedCount in
+		var lhs = buffer.baseAddress!
+		var rhs = lhs + buffer.count
+		do {
+		  for element in self {
+			if try belongsInSecondCollection(element) {
+			  rhs -= 1
+			  rhs.initialize(to: element)
+			} else {
+			  lhs.initialize(to: element)
+			  lhs += 1
+			}
+		  }
+		  
+		  let rhsIndex = rhs - buffer.baseAddress!
+		  buffer[rhsIndex...].reverse()
+		  initializedCount = buffer.count
+		  
+		  midPoint = rhsIndex
+		} catch {
+		  let lhsCount = lhs - buffer.baseAddress!
+		  let rhsCount = (buffer.baseAddress! + buffer.count) - rhs
+		  buffer.baseAddress!.deinitialize(count: lhsCount)
+		  rhs.deinitialize(count: rhsCount)
+		  throw error
+		}
+	  })
+	
+	let collections = elements.partitioned(upTo: midPoint)
+	return _tupleMap(collections, { Array($0) })
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// partitioned(upTo:)
+//===----------------------------------------------------------------------===//
+
+extension Collection {
+  /// Splits the receiving collection into two at the specified index
+  /// - Parameter index: The index within the receiver to split the collection
+  /// - Returns: A tuple with the first and second parts of the receiving
+  /// collection after splitting it
+  /// - Note: The first subsequence in the returned tuple does *not* include
+  /// the element at `index`. That element is in the second subsequence.
+  /// - Complexity: O(*1*)
+  @inlinable
+  public func partitioned(upTo index: Index) -> (SubSequence, SubSequence) {
+	return (
+	  self[self.startIndex..<index],
+	  self[index..<self.endIndex]
+	)
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// _tupleMap(_:_:)
+//===----------------------------------------------------------------------===//
+
+/// Returns a tuple containing the results of mapping the given closure over
+/// each of the tuple’s elements.
+/// - Parameters:
+///   - x: The tuple to transform
+///   - transform: A mapping closure. `transform` accepts an element of this
+///   sequence as its parameter and returns a transformed
+/// - Returns: A tuple containing the transformed elements of this tuple.
+@usableFromInline
+internal func _tupleMap<T, U>(
+  _ x: (T, T),
+  _ transform: (T) throws -> U
+) rethrows -> (U, U) {
+  return (
+	try transform(x.0),
+	try transform(x.1)
+  )
+}
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Read more about the package, and the intent behind it, in the [announcement on s
 #### Subsetting operations
 
 - [`compacted()`](https://github.com/apple/swift-algorithms/blob/main/Guides/Compacted.md): Drops the `nil`s from a sequence or collection, unwrapping the remaining elements.
+- [`partitioned(_:)`](https://github.com/apple/swift-algorithms/blob/main/Guides/Partition.md): Returns the elements in a sequence or collection that do and not match a given predciate.
 - [`randomSample(count:)`, `randomSample(count:using:)`](https://github.com/apple/swift-algorithms/blob/main/Guides/RandomSampling.md): Randomly selects a specific number of elements from a collection.
 - [`randomStableSample(count:)`, `randomStableSample(count:using:)`](https://github.com/apple/swift-algorithms/blob/main/Guides/RandomSampling.md): Randomly selects a specific number of elements from a collection, preserving their original relative order.
 - [`striding(by:)`](https://github.com/apple/swift-algorithms/blob/main/Guides/Stride.md): Returns every nth element of a collection.
--- a/Tests/SwiftAlgorithmsTests/PartitionTests.swift
+++ b/Tests/SwiftAlgorithmsTests/PartitionTests.swift
@@ -133,4 +133,63 @@ final class PartitionTests: XCTestCase {
       }
     }
   }
+  
+  func testPartitionedWithEmptyInput() {
+    let input: [Int] = []
+    
+    let s0 = input.partitioned({ _ in return true })
+    
+    XCTAssertTrue(s0.0.isEmpty)
+    XCTAssertTrue(s0.1.isEmpty)
+  }
+  
+  /// Test the example given in the `partitioned(_:)` documentation
+  func testPartitionedExample() throws {
+    let cast = ["Vivien", "Marlon", "Kim", "Karl"]
+    let (longNames, shortNames) = cast.partitioned({ $0.count < 5 })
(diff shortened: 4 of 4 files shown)


mdznr:
I made the closure `belongsInSecondCollection`, which is consistent with the other partitioned functions, but in terms of the return values, feels backwards (`(second, first)`).

mdznr:
I ran some benchmarks using the awesome [swift-collections-benchmark](https://github.com/apple/swift-collections-benchmark) package, as [suggested](https://github.com/apple/swift-algorithms/pull/151#issuecomment-880046684) by @timvermeulen:

chart

The output does confirm that using `partitioned(_:)` is faster than calling `filter(_:)` twice.

Using the `Collection`-based implementation with the fixed buffer size is faster than the `Sequence`-base implementation. However, the function’s overhead makes it slightly slower for collections fewer than 8 elements. For that reason, we could check `count < 8` in the start of the `Collection` implementation and conditionally run the `Sequence`-based approach, which is slightly faster.

I was initially surprised `partitioned(_:)` wasn’t significantly faster than calling `filter(_:)` twice, though. With more experimentation, I learned how much the overall cost depends on the cost of the closure’s evaluation. The more costly the closure, the more valuable it was to avoid calling the closure twice (obviously). In very simple cases, the cost of evaluating the closure is extremely insignificant compared to the cost of adding each element to the output collection. However, in all cases, it is still faster to use `partitioned(_:)` than calling `filter(_:)` twice.

Using a slighty more expensive closure yielded these results:

chart

<details>
<summary>Benchmarking code/details</summary>
Simple closure test:

```swift
benchmark.addSimple(
  title: "Filter × 2",
  input: [Int].self
) { input in
  blackHole(input.filter({ $0.isMultiple(of: 3) }))
  blackHole(input.filter({ !$0.isMultiple(of: 3) }))
}

benchmark.addSimple(
  title: "Partitioned (Sequence)",
  input: [Int].self
) { input in
  blackHole(input._partitioned({ $0.isMultiple(of: 3) }))
}

benchmark.addSimple(
  title: "Partitioned (Collection)",
  input: [Int].self
) { input in
  blackHole(input.partitioned({ $0.isMultiple(of: 3) }))
}
```

More expensive closure test:

```swift
let multiples: [Int] = [1, 3, 5, 7]

benchmark.addSimple(
  title: "Filter × 2",
  input: [Int].self
) { input in
  blackHole(input.__filter({ int in multiples.allSatisfy({ int.isMultiple(of: $0) }) }))
  blackHole(input.__filter({ int in !multiples.allSatisfy({ int.isMultiple(of: $0) }) }))
}

benchmark.addSimple(
  title: "Partitioned (Sequence)",
  input: [Int].self
) { input in
  blackHole(input._partitioned({ int in multiples.allSatisfy({ int.isMultiple(of: $0) }) }))
}

benchmark.addSimple(
  title: "Partitioned (Collection)",
  input: [Int].self
) { input in
  blackHole(input.partitioned({ int in multiples.allSatisfy({ int.isMultiple(of: $0) }) }))
}
```

_All tests run on iMac Pro 3.2 GHz 8-Core Intel Xeon W; 32 GB 2666 MHz DDR4; macOS 11.3 (20E232); Apple Swift version 5.4.2 (swiftlang-1205.0.28.2 clang-1205.0.19.57)_

</details>

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+    // than 8 elements. Use the simple `Sequence`-based implementation instead.
+    // This constant was determined using benchmarking. More information:
+    // https://github.com/apple/swift-algorithms/pull/152#issuecomment-887130149
+    if count < 8 {
+      return try _partitioned(belongsInSecondCollection)
+    }
+    
+    // Inside of the `initializer` closure, we set what the actual mid-point is.
+    // We will use this to partitioned the single array into two in constant time.
+    var midPoint: Int = 0
```
mdznr, line 296: Curious to hear if there’s feedback for a better name for this local variable

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+    // right elements on the right side in reverse order, we need to reverse
+    // them back to the original order at the end.
+    
+    let count = self.count
+    
+    // The overhead of this implementation isn’t worth it for collections fewer
+    // than 8 elements. Use the simple `Sequence`-based implementation instead.
+    // This constant was determined using benchmarking. More information:
+    // https://github.com/apple/swift-algorithms/pull/152#issuecomment-887130149
+    if count < 8 {
```
mdznr, line 290: Curious to hear feedback about how/where to define these kinds of constants.

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+  /// `RandomAccessCollection`.
+  @inlinable
+  public func partitioned(
+    _ belongsInSecondCollection: (Element) throws -> Bool
+  ) rethrows -> ([Element], [Element]) {
+    return try _partitioned(belongsInSecondCollection)
+  }
+  
+  @inlinable
+  public func _partitioned(
```
mdznr, line 248: I’m curious for feedback on how to best let this implementation be made available internally with all the benefits of having it inlined. The [`filter(_:)` implementation in the standard library](https://github.com/apple/swift/blob/7123d2614b5f222d03b3762cb110d27a9dd98e24/stdlib/public/core/Sequence.swift#L667) uses `@_transparent`, but it seems like this [might not be appropriate](https://github.com/apple/swift/blob/main/docs/TransparentAttr.md#when-should-you-use-_transparent) for use in a module like this.

Guides/Partition.md
```diff
@@ -69,11 +95,17 @@ extension Collection {
         where belongsInSecondPartition: (Element) throws -> Bool
     ) rethrows -> Index
 }
+
+extension Sequence {
+	public func bifurcate(
```
CTMacUser, line 100: You forgot to rename this instance in the documentation.

Sources/Algorithms/Partition.swift
timvermeulen: Using a literal alongside a comment on what it represents is fine. I'm not fully convinced we need this branch though, small cases are going to be very fast either way and it adds a bit of complexity. It also currently causes all tests to use the `Sequence` implementation 😅

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+              rhs -= 1
+              rhs.initialize(to: element)
+            } else {
+              lhs.initialize(to: element)
+              lhs += 1
+            }
+          }
+          
+          let rhsIndex = rhs - buffer.baseAddress!
+          buffer[rhsIndex...].reverse()
```
timvermeulen, line 315: This reverse operation can be deferred until these elements are stored in a newly allocated array, that might speed things up slightly.

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+  /// the element at `index`. That element is in the second subsequence.
+  /// - Complexity: O(*1*)
+  @inlinable
+  public func partitioned(upTo index: Index) -> (SubSequence, SubSequence) {
+    return (
+      self[self.startIndex..<index],
+      self[index..<self.endIndex]
+    )
+  }
+}
```
timvermeulen, line 352: This operation is so simple that I don't think there's a point in including it here, it's not a particularly common operation and `(self[..<index], self[index...])` isn't too verbose.

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+@usableFromInline
+internal func _tupleMap<T, U>(
+  _ x: (T, T),
+  _ transform: (T) throws -> U
+) rethrows -> (U, U) {
+  return (
+    try transform(x.0),
+    try transform(x.1)
+  )
+}
```
timvermeulen, line 374: Same here, IMO this does not benefit the readability of the code currently

timvermeulen (commented):
Those graphs look good! It's nice to see that the added complexity seems to be worth it for most sizes. I think it'd be useful to benchmark another version that returns a `(ArraySlice, ArraySlice)` pair (or even `(ArraySlice, ReversedCollection<ArraySlice>)`) just so we can see how much performance we're missing out on by allocating two new arrays.

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+  ) rethrows -> ([Element], [Element]) {
+    return try _partitioned(belongsInSecondCollection)
+  }
+  
+  @inlinable
+  public func _partitioned(
+    _ belongsInSecondCollection: (Element) throws -> Bool
+  ) rethrows -> ([Element], [Element]) {
+    var lhs = ContiguousArray<Element>()
+    var rhs = ContiguousArray<Element>()
```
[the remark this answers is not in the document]
mdznr: `ContiguousArray` and `Array` seem to perform about the same:
chart

I’ve changed it to `Array`

Sources/Algorithms/Partition.swift
mdznr: When would the reversal happen? If removing line 315 here and instead reversing the `ArraySlice` right before its conversion to an `Array`, it makes it slower.

Before:
```swift
          buffer[rhsIndex...].reverse()
          // …
        } catch {
          // …
        }
      })
    
    let lhs = elements[..<midPoint]
    let rhs = elements[midPoint...]
    return (
      Array(lhs),
      Array(rhs)
    )
```

After:
```swift
    let lhs = elements[..<midPoint]
    let rhs = elements[midPoint...].reversed()
    return (
      Array(lhs),
      Array(rhs)
    )
```

chart

mdznr:
> Those graphs look good! It's nice to see that the added complexity seems to be worth it for most sizes. I think it'd be useful to benchmark another version that returns a `(ArraySlice, ArraySlice)` pair (or even `(ArraySlice, ReversedCollection<ArraySlice>)`) just so we can see how much performance we're missing out on by allocating two new arrays.

chart

<details>
<summary>Closeup of the two slice variants</summary>

chart
</details>

<details>
<summary>Code</summary>

```swift
extension Collection {
  @inlinable
  public func partitionedA(
    _ belongsInSecondCollection: (Element) throws -> Bool
  ) rethrows -> ([Element], [Element]) {
    guard !self.isEmpty else {
      return ([], [])
    }
    
    // Since `RandomAccessCollection`s have known sizes (access to `count` is
    // constant time, O(1)), we can allocate one array of size `self.count`,
    // then insert items at the beginning or end of that contiguous block. This
    // way, we don’t have to do any dynamic array resizing. Since we insert the
    // right elements on the right side in reverse order, we need to reverse
    // them back to the original order at the end.
    
    let count = self.count
    
    // Inside of the `initializer` closure, we set what the actual mid-point is.
    // We will use this to partitioned the single array into two in constant time.
    var midPoint: Int = 0
    
    let elements = try [Element](
      unsafeUninitializedCapacity: count,
      initializingWith: { buffer, initializedCount in
        var lhs = buffer.baseAddress!
        var rhs = lhs + buffer.count
        do {
          for element in self {
            if try belongsInSecondCollection(element) {
              rhs -= 1
              rhs.initialize(to: element)
            } else {
              lhs.initialize(to: element)
              lhs += 1
            }
          }
          
          let rhsIndex = rhs - buffer.baseAddress!
          buffer[rhsIndex...].reverse()
          initializedCount = buffer.count
          
          midPoint = rhsIndex
        } catch {
          let lhsCount = lhs - buffer.baseAddress!
          let rhsCount = (buffer.baseAddress! + buffer.count) - rhs
          buffer.baseAddress!.deinitialize(count: lhsCount)
          rhs.deinitialize(count: rhsCount)
          throw error
        }
      })
    
    let lhs = elements[..<midPoint]
    let rhs = elements[midPoint...]
    return (
      Array(lhs),
      Array(rhs)
    )
  }
}

extension Collection {
  @inlinable
  public func partitionedB(
    _ belongsInSecondCollection: (Element) throws -> Bool
  ) rethrows -> (ArraySlice<Element>, ArraySlice<Element>) {
    guard !self.isEmpty else {
      return ([], [])
    }
    
    // Since `RandomAccessCollection`s have known sizes (access to `count` is
    // constant time, O(1)), we can allocate one array of size `self.count`,
    // then insert items at the beginning or end of that contiguous block. This
    // way, we don’t have to do any dynamic array resizing. Since we insert the
    // right elements on the right side in reverse order, we need to reverse
    // them back to the original order at the end.
    
    let count = self.count
    
    // Inside of the `initializer` closure, we set what the actual mid-point is.
    // We will use this to partitioned the single array into two in constant time.
    var midPoint: Int = 0
    
    let elements = try [Element](
      unsafeUninitializedCapacity: count,
      initializingWith: { buffer, initializedCount in
        var lhs = buffer.baseAddress!
        var rhs = lhs + buffer.count
        do {
          for element in self {
            if try belongsInSecondCollection(element) {
              rhs -= 1
              rhs.initialize(to: element)
            } else {
              lhs.initialize(to: element)
              lhs += 1
            }
          }
          
          let rhsIndex = rhs - buffer.baseAddress!
          buffer[rhsIndex...].reverse()
          initializedCount = buffer.count
          
          midPoint = rhsIndex
        } catch {
          let lhsCount = lhs - buffer.baseAddress!
          let rhsCount = (buffer.baseAddress! + buffer.count) - rhs
          buffer.baseAddress!.deinitialize(count: lhsCount)
          rhs.deinitialize(count: rhsCount)
          throw error
        }
      })
    
    let lhs = elements[..<midPoint]
    let rhs = elements[midPoint...]
    return (lhs, rhs)
  }
}

extension Collection {
  @inlinable
  public func partitionedC(
    _ belongsInSecondCollection: (Element) throws -> Bool
  ) rethrows -> (ArraySlice<Element>, ReversedCollection<ArraySlice<Element>>) {
    guard !self.isEmpty else {
      let emptyArraySlice = [Element]()[0...]
      return (
        emptyArraySlice,
        emptyArraySlice.reversed()
      )
    }
    
    // Since `RandomAccessCollection`s have known sizes (access to `count` is
    // constant time, O(1)), we can allocate one array of size `self.count`,
    // then insert items at the beginning or end of that contiguous block. This
    // way, we don’t have to do any dynamic array resizing. Since we insert the
    // right elements on the right side in reverse order, we need to reverse
    // them back to the original order at the end.
    
    let count = self.count
    
    // Inside of the `initializer` closure, we set what the actual mid-point is.
    // We will use this to partitioned the single array into two in constant time.
    var midPoint: Int = 0
    
    let elements = try [Element](
      unsafeUninitializedCapacity: count,
      initializingWith: { buffer, initializedCount in
        var lhs = buffer.baseAddress!
        var rhs = lhs + buffer.count
        do {
          for element in self {
            if try belongsInSecondCollection(element) {
              rhs -= 1
              rhs.initialize(to: element)
            } else {
              lhs.initialize(to: element)
              lhs += 1
            }
          }
          
          let rhsIndex = rhs - buffer.baseAddress!
          initializedCount = buffer.count
          
          midPoint = rhsIndex
        } catch {
          let lhsCount = lhs - buffer.baseAddress!
          let rhsCount = (buffer.baseAddress! + buffer.count) - rhs
          buffer.baseAddress!.deinitialize(count: lhsCount)
          rhs.deinitialize(count: rhsCount)
          throw error
        }
      })
    
    let lhs = elements[..<midPoint]
    let rhs = elements[midPoint...]
    return (lhs, rhs.reversed())
  }
}
```

```swift
benchmark.addSimple(
  title: "Array, Array",
  input: [Int].self
) { input in
  blackHole(input.partitionedA({
    $0.isMultiple(of: 2)
  }))
}

benchmark.addSimple(
  title: "ArraySlice, ArraySlice",
  input: [Int].self
) { input in
  blackHole(input.partitionedB({
    $0.isMultiple(of: 2)
  }))
}

benchmark.addSimple(
  title: "ArraySlice, ReversedCollection<ArraySlice>",
  input: [Int].self
) { input in
  blackHole(input.partitionedC({
    $0.isMultiple(of: 2)
  }))
}
```
</details>

I’m a bit surprised that the `(Array, Array)` implementation was actually faster in many cases (from a few hundred to a couple hundred thousand elements). Thinking it was a fluke, I’ve re-run this several times and continue to get similar results. I’m not sure yet why that could be.

I wish there were a clear best implementation from a performance point of view. However, since the performance for the non-`Array` return values weren’t better in all cases, I would say the tradeoff of having non-`Array` types as the return values aren’t worth it, as it does expose some implementation details and would make it difficult to change the implementation (and possibly the signature) later without it being a non-backwards-compatible breaking change.

timvermeulen:
> I’m a bit surprised that the `(Array, Array)` implementation was actually faster in many cases (from a few hundred to a couple hundred thousand elements). Thinking it was a fluke, I’ve re-run this several times and continue to get similar results. I’m not sure yet why that could be.

That's interesting and indeed surprising.

> I wish there were a clear best implementation from a performance point of view. However, since the performance for the non-`Array` return values weren’t better in all cases, I would say the tradeoff of having non-`Array` types as the return values aren’t worth it, as it does expose some implementation details and would make it difficult to change the implementation (and possibly the signature) later without it being a non-backwards-compatible breaking change.

I completely agree with your conclusions here. I'd still be interested to see how returning `(Array(lhs), Array(rhs.reversed()))` rather than reversing the right side in-place could improve the `(Array, Array)` version even more, but judging by the tiny difference between the two slice versions it probably won't make a huge difference.

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,154 @@ extension Collection {
...
+    
+    for element in self {
+      if try belongsInSecondCollection(element) {
+        rhs.append(element)
+      } else {
+        lhs.append(element)
+      }
+    }
+    
+    return (
```
LucianoPAlmeida, line 262: Just curious if there's a reason for we not just `return (lhs, rhs)`?

mdznr: Nope. I just hadn’t pushed the latest changes. I was playing around with the performance implications earlier of `ContiguousArray` and `Array` and this was just a holdover from that.

mdznr:
> > I wish there were a clear best implementation from a performance point of view. However, since the performance for the non-`Array` return values weren’t better in all cases, I would say the tradeoff of having non-`Array` types as the return values aren’t worth it, as it does expose some implementation details and would make it difficult to change the implementation (and possibly the signature) later without it being a non-backwards-compatible breaking change.
> 
> I completely agree with your conclusions here. I'd still be interested to see how returning `(Array(lhs), Array(rhs.reversed()))` rather than reversing the right side in-place could improve the `(Array, Array)` version even more, but judging by the tiny difference between the two slice versions it probably won't make a huge difference.

I think if I’m following you correctly, that should be the same as the [test I ran earlier](https://github.com/apple/swift-algorithms/pull/152#discussion_r691536282):

> When would the reversal happen? If removing line 315 here and instead reversing the `ArraySlice` right before its conversion to an `Array`, it makes it slower.
> chart

timvermeulen:
> I think if I’m following you correctly, that should be the same as the [test I ran earlier](https://github.com/apple/swift-algorithms/pull/152#discussion_r691536282):

I missed that, my bad. Looks like `ReversedCollection` is just too slow to make this worthwhile.

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,171 @@ extension Collection {
...
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the sequence.
+  ///
+  /// - Note: This algorithm performs a bit slower than the same algorithm on
+  /// `RandomAccessCollection` since the size of the sequence is unknown, unlike
+  /// `RandomAccessCollection`.
+  @inlinable
+  public func partitioned(
+    _ belongsInSecondCollection: (Element) throws -> Bool
+  ) rethrows -> ([Element], [Element]) {
```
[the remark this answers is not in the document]
mdznr: I was thinking maybe `excluded` and `included`, respectively. While they would be accurate, they don’t really sound great alongside the `belongsInSecondCollection` parameter name. I think keeping the parameter name and behavior consistent across the partition* family of functions is important.

Guides/Partition.md
```diff
@@ -69,11 +83,17 @@ extension Collection {
         where belongsInSecondPartition: (Element) throws -> Bool
     ) rethrows -> Index
 }
+
+extension Sequence {
+	public func partitioned(
+	  _ belongsInSecondCollection: (Element) throws -> Bool
+	) rethrows -> ([Element], [Element])
```
xwu, line 90:
```suggestion
    public func partitioned(
        _ belongsInSecondCollection: (Element) throws -> Bool
    ) rethrows -> ([Element], [Element])
```

Sources/Algorithms/Partition.swift
```diff
@@ -204,3 +204,144 @@ extension Collection {
...
+    var lhs = Array<Element>()
+    var rhs = Array<Element>()
```
xwu, line 245: Nit for consistency with the rest of the code here:

```suggestion
    var lhs = [Element]()
    var rhs = [Element]()
```

Sources/Algorithms/Partition.swift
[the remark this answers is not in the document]
timvermeulen: We discussed this internally and we agree that the combination of `belongsInSecondCollection` in combination with `included`/`excluded` would be odd. However, note that the closure parameter of the mutating versions is called `belongsInSecondPartition` primarily because it's the only way to specify which side of the collection corresponds to the elements that produce `true`. Here we have tuple labels to work with though, which I think serves this purpose a lot better than the parameter name. How does that sound? `falseElements:` and `trueElements:` are the labels we came up with (in that order, to still stay as consistent as we can with the mutating versions). 

Let's also add an external `by:` label here, the mutating versions don't have one but there's no reason why they shouldn't. We'll probably want to change those later to include it as well.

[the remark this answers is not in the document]
mdznr: I think that’s probably the best thing we can do in this scenario. Thanks Tim and team!

natecook1000:
@swift-ci Please test

PR #153, opened 2021-07-23
Add `commonPrefix(with:)`, `commonSuffix(with:)`

timvermeulen:
Find the common prefix/suffix of two sequences or collections.

```swift
extension Sequence {
  public func commonPrefix<Other: Sequence>(
    with other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> [Element]
}

extension Sequence where Element: Equatable {
  public func commonPrefix<Other: Sequence>(
    with other: Other
  ) -> CommonPrefix<Self, Other> where Other.Element == Element
}

extension LazySequenceProtocol {
  public func commonPrefix<Other: Sequence>(
    with other: Other,
    by areEquivalent: @escaping (Element, Other.Element) -> Bool
  ) -> CommonPrefix<Self, Other>
}

extension Collection {
  public func commonPrefix<Other: Sequence>(
    with other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> SubSequence
}

extension Collection where Element: Equatable {
  public func commonPrefix<Other: Sequence>(
    with other: Other
  ) -> SubSequence where Other.Element == Element
}

extension LazyCollectionProtocol {
  public func commonPrefix<Other: Sequence>(
    with other: Other,
    by areEquivalent: @escaping (Element, Other.Element) -> Bool
  ) -> CommonPrefix<Self, Other>
}

extension LazyCollectionProtocol where Element: Equatable {
  public func commonPrefix<Other: Sequence>(
    with other: Other
  ) -> CommonPrefix<Self, Other> where Other.Element == Element
}

  public func commonSuffix<Other: BidirectionalCollection>(
    with other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> SubSequence
}

  public func commonSuffix<Other: BidirectionalCollection>(
    with other: Other
  ) -> SubSequence where Other.Element == Element
}

extension Collection {
  public func endsOfCommonPrefix<Other: Collection>(
    with other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> (Index, Other.Index)
}

extension Collection where Element: Equatable {
  public func endsOfCommonPrefix<Other: Collection>(
    with other: Other
  ) -> (Index, Other.Index) where Other.Element == Element
}

  public func startsOfCommonSuffix<Other: BidirectionalCollection>(
    with other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> (Index, Other.Index)
}

  public func startsOfCommonSuffix<Other: BidirectionalCollection>(
    with other: Other
  ) -> (Index, Other.Index) where Other.Element == Element
}
```

### Checklist

### DIFF
--- a/Sources/Algorithms/CommonPrefix.swift
+++ b/Sources/Algorithms/CommonPrefix.swift
@@ -1,0 +1,350 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Algorithms open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public struct CommonPrefix<Base: Sequence, Other: Sequence> {
+  @usableFromInline
+  internal let base: Base
+  
+  @usableFromInline
+  internal let other: Other
+  
+  @usableFromInline
+  internal let areEquivalent: (Base.Element, Other.Element) -> Bool
+  
+  @inlinable
+  internal init(
+    base: Base,
+    other: Other,
+    areEquivalent: @escaping (Base.Element, Other.Element) -> Bool
+  ) {
+    self.base = base
+    self.other = other
+    self.areEquivalent = areEquivalent
+  }
+}
+
+extension CommonPrefix: Sequence {
+  public struct Iterator: IteratorProtocol {
+    @usableFromInline
+    internal var base: Base.Iterator
+    
+    @usableFromInline
+    internal var other: Other.Iterator
+    
+    @usableFromInline
+    internal let areEquivalent: (Base.Element, Other.Element) -> Bool
+    
+    @inlinable
+    internal init(
+      base: Base.Iterator,
+      other: Other.Iterator,
+      areEquivalent: @escaping (Base.Element, Other.Element) -> Bool
+    ) {
+      self.base = base
+      self.other = other
+      self.areEquivalent = areEquivalent
+    }
+    
+    public mutating func next() -> Base.Element? {
+      if let next = base.next(),
+         let otherNext = other.next(),
+         areEquivalent(next, otherNext) {
+        return next
+      } else {
+        return nil
+      }
+    }
+  }
+  
+  @inlinable
+  public func makeIterator() -> Iterator {
+    Iterator(
+      base: base.makeIterator(),
+      other: other.makeIterator(),
+      areEquivalent: areEquivalent)
+  }
+}
+
+extension CommonPrefix: Collection where Base: Collection, Other: Collection {
+  public struct Index {
+    @usableFromInline
+    internal let base: Base.Index
+    
+    @usableFromInline
+    internal let other: Other.Index
+
+    @inlinable
+    internal init(base: Base.Index, other: Other.Index) {
+      self.base = base
+      self.other = other
+    }
+  }
+  
+  @inlinable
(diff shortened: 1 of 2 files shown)


kylemacomber:
I think it's more natural to not pluralize of `start` and `end`, i.e. just have `endOfCommonPrefix`, `startOfCommonSuffix`

timvermeulen:
> I think it's more natural to not pluralize of `start` and `end`, i.e. just have `endOfCommonPrefix`, `startOfCommonSuffix`

Fair point. I think I went with this because we already have `endOfPrefix(where:)` and `startOfSuffix(where:)` which both return a single index, but that's probably fine.

Sources/Algorithms/CommonPrefix.swift
```diff
@@ -0,0 +1,350 @@
...
+    
+    public mutating func next() -> Base.Element? {
+      if let next = base.next(),
+         let otherNext = other.next(),
+         areEquivalent(next, otherNext) {
+        return next
+      } else {
+        return nil
+      }
+    }
```
CTMacUser, line 64: I think the iterator needs to add a `done` flag to its state.  For instance, if we do `"1234A56"` vs. `"1234B56"`, and we iterate past the first `nil`, we'll get a `"5"`, instead of `nil` forevermore.

Sources/Algorithms/CommonPrefix.swift
```diff
@@ -0,0 +1,350 @@
...
+//===----------------------------------------------------------------------===//
+
+extension LazySequenceProtocol {
+  @inlinable
+  public func commonPrefix<Other: Sequence>(
+    with other: Other,
+    by areEquivalent: @escaping (Element, Other.Element) -> Bool
+  ) -> CommonPrefix<Self, Other> {
+    CommonPrefix(base: self, other: other, areEquivalent: areEquivalent)
+  }
```
CTMacUser, line 193: Shouldn't these be using `Elements` and `elements` instead of `Self` and `self`?

Sources/Algorithms/CommonPrefix.swift
```diff
@@ -0,0 +1,350 @@
...
+//===----------------------------------------------------------------------===//
+
+extension LazyCollectionProtocol {
+  @inlinable
+  public func commonPrefix<Other: Sequence>(
+    with other: Other,
+    by areEquivalent: @escaping (Element, Other.Element) -> Bool
+  ) -> CommonPrefix<Self, Other> {
+    CommonPrefix(base: self, other: other, areEquivalent: areEquivalent)
+  }
```
CTMacUser, line 242: Same as in the `LazySequenceProtocol` section: shouldn't these be using `Elements` and `elements` instead of `Self` and `self`?

CTMacUser:
Don't `endsOfCommonPrefix(with:)` and `startssOfCommonSuffix(with:)` overlap with `diverges(from:)` and `converges(with:)` from #37 ?

Sources/Algorithms/CommonPrefix.swift
timvermeulen: That's a way to do it, but this is intentional. If you wrap `elements` instead of `self` then the type no longer contains any sign of laziness, so you'll need to wrap it in `LazySequence<...>` again to add the laziness back. But by wrapping `self`, all we need is for `CommonPrefix` to conditionally conform to `LazySequenceProtocol` when the (first) base collection does, in order to propagate the laziness.

`CommonPrefix` also can't unconditionally conform to `LazySequenceProtocol` because it's also returned by the `Sequence` overloads, so in a non-lazy context.

PR #154, opened 2021-07-23
Add `firstRange(of:)`, `lastRange(of:)`, `contains(_:)`

timvermeulen:
Find the location of a subsequence in a collection.

* `firstRange(of:)`
* `lastRange(of:)`
* `contains(_:)`

```swift
extension Collection {
  public func firstRange<Other: Collection>(
    of other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> Range<Index>?
}

extension Collection where Element: Equatable {
  public func firstRange<Other: Collection>(of other: Other) -> Range<Index>?
}

  public func lastRange<Other: BidirectionalCollection>(
    of other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> Range<Index>?
}

  public func lastRange<Other: BidirectionalCollection>(
    of other: Other
  ) -> Range<Index>? where Other.Element == Element
}

extension Collection {
  public func contains<Other: Collection>(
    _ other: Other,
    by areEquivalent: (Element, Other.Element) throws -> Bool
  ) rethrows -> Bool
}

extension Collection where Element: Equatable {
  public func contains<Other: Collection>(
    _ other: Other
  ) -> Bool where Other.Element == Element
}
```

### Checklist

### DIFF
--- a/Sources/Algorithms/FirstRange.swift
+++ b/Sources/Algorithms/FirstRange.swift
@@ -14,6 +14,26 @@
 //===----------------------------------------------------------------------===//
 
 extension Collection {
+  /// Returns the first index range where this collection contains the same
+  /// elements as another collection in the same order, according to the given
+  /// equivalence function.
+  ///
+  ///     let string = "foo, bar, foo, bar"
+  ///     if let range = string.firstRange(of: "bar", by: ==) {
+  ///         print(string[..<range.lowerBound]) // "foo, "
+  ///         print(string[range])               // "bar"
+  ///         print(string[range.upperBound...]) // ", foo, bar"
+  ///     }
+  ///
+  /// - Parameters:
+  ///   - other: The collection to search for.
+  ///   - areEquivalent: A predicate that returns true if its two arguments are
+  ///     equivalent; otherwise, false.
+  /// - Returns: The first index range `r` such that `self[r]` equals `other`
+  ///   according to `areEquivalent`.
+  ///
+  /// - Complexity: O(*n* \* *m*), where *n* is the length of this collection
+  ///   and *m* is the length of the collection being searched for.
   @inlinable
   public func firstRange<Other: Collection>(
     of other: Other,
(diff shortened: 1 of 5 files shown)


mdznr:
There could probably be optimizations based on `count`.

For example, if `other`’s `count` is larger than `self`, then there’s no point in iterating through `self` since it could never contain a larger collection as a subsequence. That’s just a base case that could be checked up front. However, this can also happen in the middle of the algorithm. For example, if we had a `self.count` of 20, and an `other.count` of 5, if we ever get a starting index (`matchStart`) past the 15th element, we can then just return `nil` since we can’t fit a subsequence of 5 with only 4 elements remaining.

glessard:
@mdznr Such an optimization would only work for a `RandomAccessCollection`, and these are not specialized for those indexing semantics (yet.)

Sources/Algorithms/FirstRange.swift
```diff
@@ -0,0 +1,133 @@
...
+// lastRange(of:)
+//===----------------------------------------------------------------------===//
+
+extension BidirectionalCollection {
+  @inlinable
+  public func lastRange<Other: BidirectionalCollection>(
+    of other: Other,
+    by areEquivalent: (Element, Other.Element) throws -> Bool
+  ) rethrows -> Range<Index>? {
+    var searchEnd = endIndex
```
CTMacUser, line 71: I originally thought this method would just use wrapping `ReversedCollection`s, call `firstRange(of:)`, and counter-convert the returned indexes.  I don't know how much of an improvement this customized code is.  (The downside is that improvements to `firstRange(of:)` aren't automatically inherited.)  Maybe we could use benchmarking?

timvermeulen: That would have been nice, but `ReversedCollection` unfortunately does an index decrement inside its [subscript](https://github.com/apple/swift/blob/main/stdlib/public/core/Reverse.swift#L246-L249), which makes it unsuitable for such purposes.

natecook1000:
This functionality is handled by the string-processing package.

PR #157, opened 2021-07-23
Add `paddingStart(with:toCount:)` and `paddingEnd(with:toCount:)` operations to `RangeReplaceableCollection`.

Sameesunkaria:
### Description

Introduces "padding" methods on `RangeReplaceableCollection` to expand the collection's length to the desired length by repeating a padding element at the start or at the end. A padding operation is often performed on strings or data to achieve a fixed length output.

This change adds two new operations, `paddingStart` and `paddingEnd` and the corresponding mutating variants `padStart` and `padEnd`. Apple's `Foundation` has a similar padding API [`padding(toLength:withPad:startingAt:)`](https://developer.apple.com/documentation/foundation/nsstring/1416395-padding) but, it is only available on `NSStrings`. `paddingStart` and `paddingEnd` operate on any Swift `RangeReplaceableCollection`. There also seems to be [some interest for adding a padding method for arrays on the Swift forums](https://forums.swift.org/t/padding-arrays/41041), the two newly added methods should address those demands.

A couple of example use-cases are shown below:

```swift
let paddedNumber = "96".paddingStart(with: "0", toCount: 4)
// paddedNumber == "0096"

let data = Data([0xde, 0xad, 0xbe, 0xef])
let paddedData = data.paddingEnd(with: 0, toCount: 16)
// paddedData contains 16 bytes
```

### Detailed Design

Two new operations, `paddingStart` and `paddingEnd` have been introduced to `RangeReplaceableCollection`.

```swift
extension RangeReplaceableCollection {
    public func paddingStart(
        with element: Element,
        toCount paddedCount: Int
    ) -> Self
  
    public func paddingEnd(
        with element: Element,
        toCount paddedCount: Int
    ) -> Self
}
```

Each of the two operations also have a mutating varient.

```swift
extension RangeReplaceableCollection {
    public mutating func padStart(
        with element: Element,
        toCount paddedCount: Int)
      
    public mutating func padEnd(
        with element: Element,
        toCount paddedCount: Int)
}
```

Padding a collection expands the collection to the desired length by repeating the padding element at the start or the end. In case the collection's length is greater than the `paddedCount`, the collection is preserved.

### Documentation Plan

The documentation is WIP both in code (documentation comments) and the markdown guide. I am looking for additional feedback from the community whether this change belongs in swift-algorithms and if the API or implementation needs to be changed or improved. Complete documentation will be added once the API is finalized. 

### Test Plan

Unit tests have been added for the newly introduced collection types.

### Source Impact

This is a purely additive change.

### Checklist


timvermeulen:
Thanks for your contribution, @Sameesunkaria! This looks excellent, and it's a great fit for this package.

With regard to the implementation strategy, I think the right approach is to make this a wrapper around a `Chain2` instance. It's totally fine (and even encouraged!) to have a dependency on other functionality from this package. We already do that in several places, a good example is how `FiniteCycle` is a thin wrapper around a `Product2` instance.

> * An eager implementation on `RangeReplaceableCollection`.

Mutating variants on `RangeReplaceableColleciton` would be great to have for situations where you're working with an `Array` / `Data` / `String` instance and want to keep using it, rather than a custom wrapper type that doesn't have the same capabilities. It might even make sense to add non-mutating variants on `RangeReplaceableCollection` as well, although that's not as obvious to me since the wrapper type will have better performance. This package provides both mutating and non-mutating ways of [trimming](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Trim.swift) a collection, it's worth checking that out if you haven't.

For symmetry with the `trimmingPrefix` / `trimmingSuffix` methods it could make sense to name these `paddingPrefix` / `paddingSuffix` (and mutating versions `padPrefix` / `padSuffix`). What do you think about that?

Sameesunkaria:
>It's totally fine (and even encouraged!) to have a dependency on other functionality from this package.

Thanks for pointing that out. Having a wrapper around the existing collections would significantly simplify the implementation.

>It might even make sense to add non-mutating variants on `RangeReplaceableCollection` as well, although that's not as obvious to me since the wrapper type will have better performance.

In an ideal world `String(str.lazyPaddingPrefix(with: "0", toCount: 10))` and `str.eagerPaddingPrefix(with: "0", toCount: 10)` should perform the same right?

<hr>

I was also inspired by some of your other PR open currently, it may be beneficial to take another sequence for the padding elements:

```swift
extension Collection {
    func paddedPrefix<Other: Sequence>(
        with other: Other,
        toCount paddingLength: Int
    ) -> PaddedPrefix<Self, Other> where Other.Element == Element
}
```

What do you think about it? 

There is the obvious question about where should the other collection start. One might not want the `Other` sequence to start at the beginning based on the count of the number of elements, similar to Apple's [`padding(toLength:withPad:startingAt:)`](https://developer.apple.com/documentation/foundation/nsstring/1416395-padding)

timvermeulen:
> > It might even make sense to add non-mutating variants on `RangeReplaceableCollection` as well, although that's not as obvious to me since the wrapper type will have better performance.
> 
> In an ideal world `String(str.lazyPaddingPrefix(with: "0", toCount: 10))` and `str.eagerPaddingPrefix(with: "0", toCount: 10)` should perform the same right?

A padding method that constructs a new string is always going to be more expensive by necessity than a thin wrapper around a `Chain2` instance, because the former needs to allocate a new string and add every character to it while the latter is essentially a no-op.

> I was also inspired by some of your other PRs that are open currently. It may be beneficial to take another sequence for the padding elements:
> 
> ```swift
> extension Collection {
>     func paddedPrefix<Other: Sequence>(
>         with other: Other,
>         toCount paddingLength: Int
>     ) -> PaddedPrefix<Self, Other> where Other.Element == Element
> }
> ```
> 
> What do you think about it?
> 
> There is the obvious question about where should the other sequence start. One might not want the `Other` sequence to start at the beginning based on the count of the collection; similar to Apple's [`padding(toLength:withPad:startingAt:)`](https://developer.apple.com/documentation/foundation/nsstring/1416395-padding).

Great point, the Foundation API indeed lets you do more than only pad a string with a single character. We don't need to necessarily add all of the same functionality as that method, though. It's perfectly fine to only allow padding using a single element, and wait and see if people actually want to use multiple elements for padding. For what it's worth, from my limited research it seems common for other programming languages to only supply the single element version out of the box.

timvermeulen:
And maybe `prefix`/`suffix` is not the right terminology to use here because these methods don't operate on the prefix the same way `trimmingPrefix` does, so how about we go with `paddingStart`/`paddingEnd` instead?

Sameesunkaria:
>And maybe prefix/suffix is not the right terminology to use here because these methods don't operate on the prefix the same way trimmingPrefix does, so how about we go with paddingStart/paddingEnd instead?

Good point.

Sameesunkaria:
I'll benchmark a few implementations for padding on `RangeReplaceableCollection` and update the PR with the one that performs the best. I'll post the results here.

Sameesunkaria:
I performed some performance tests for the following three implementations:

```swift
extension RangeReplaceableCollection {
  func pad1(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    return Self.init(repeating: element, count: paddingCount - count) + self
  }

  func pad2(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    var paddedCollection = Self.init(repeating: element, count: paddingCount)
    let paddingEndIndex = paddedCollection.index(
      paddedCollection.startIndex,
      offsetBy:paddingCount - count)
    paddedCollection.replaceSubrange(paddedCollection.startIndex..<paddingEndIndex, with: self)
    return paddedCollection
  }

  func pad3(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    var paddedCollection = Self()
    paddedCollection.reserveCapacity(paddingCount)
    paddedCollection.append(contentsOf: repeatElement(element, count: paddingCount - count))
    paddedCollection.append(contentsOf: self)
    return paddedCollection
  }
}
```

The time taken to pad to twice the length of the original collection was considered for this benchmark. I performed 5 consecutive runs.

example using `swift-collections-benchmark`:
```swift
benchmark.addSimple(
  title: "pad1",
  input: [Int].self
) { input in
  blackHole(input.pad1(with: 0, toCount: input.count * 2))
}
```

 The results are interesting....
 
chartint

`pad1`, which I assumed would performed the worse was actually the fastest for inputs from approx. size 512 to 64k.
`pad2` and `pad3` perform similarly.

This leaves me with no clearly correct approach...

What do you think about it @timvermeulen ? Could there be something wrong with the testing strategy?

Sameesunkaria:
When I add a 4th implementation to benchmark, `pad3` ends up being equivalent or faster than `pad1`. I have repeated this a few times and got similar results. 

The four implementations:
```swift
extension RangeReplaceableCollection {
  func pad1(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    return Self.init(repeating: element, count: paddingCount - count) + self
  }

  func pad2(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    var paddedCollection = Self.init(repeating: element, count: paddingCount)
    let paddingEndIndex = paddedCollection.index(
      paddedCollection.startIndex,
      offsetBy:paddingCount - count)
    paddedCollection.replaceSubrange(paddedCollection.startIndex..<paddingEndIndex, with: self)
    return paddedCollection
  }

  func pad3(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    var paddedCollection = Self()
    paddedCollection.reserveCapacity(paddingCount)
    paddedCollection.append(contentsOf: repeatElement(element, count: paddingCount - count))
    paddedCollection.append(contentsOf: self)
    return paddedCollection
  }

  func pad4(with element: Element, toCount paddingCount: Int) -> Self {
    guard count < paddingCount else { return self }
    var paddedCollection = self
    paddedCollection.insert(contentsOf: repeatElement(element, count: paddingCount - count), at: paddedCollection.startIndex)
    return paddedCollection
  }
}
```

The data was recorded over 5 runs. 

<details>
<summary>This is how benchmark was performed:</summary>

```swift
benchmark.addSimple(
  title: "pad1",
  input: [Int].self
) { input in
  blackHole(input.pad1(with: 0, toCount: input.count * 2))
}

benchmark.addSimple(
  title: "pad2",
  input: [Int].self
) { input in
  blackHole(input.pad2(with: 0, toCount: input.count * 2))
}

benchmark.addSimple(
  title: "pad3",
  input: [Int].self
) { input in
  blackHole(input.pad3(with: 0, toCount: input.count * 2))
}

benchmark.addSimple(
  title: "pad4",
  input: [Int].self
) { input in
  blackHole(input.pad4(with: 0, toCount: input.count * 2))
}
```

</details>

`pad3` seems to be the best performer from these results.
chart

I can't explain why the results change when another case is added. Could it be something related to speculative execution?

timvermeulen:
> I can't explain why the results change when another case is added. Could it be related to some additional compiler optimizations or speculative execution?

I'm not completely sure, perhaps @lorentey could comment on this. But let's go with option 3, with the slight modification of computing `count` only once as it can be an expensive operation.

> `pad1`, which I assumed would performed the worse was actually the fastest for inputs from approx. size 512 to 64k.

I expect most padding operations to be done on collections much smaller than 512 elements, so `pad3` shows strong results regardless of which graph we look at, and its implementation is very straightfoward as well.

Sameesunkaria:
Pushed the changes for padding on RangeReplaceableCollection. Looking for feedback! 🙌

timvermeulen (commented):
This looks great, just a couple of documentation nits (some of which apply to more than one place).

Sameesunkaria:
Wow, looks like I am horrible at proof reading text 😅 
Thanks for the review ✌️

Sources/Algorithms/Pad.swift
```diff
@@ -0,0 +1,156 @@
...
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension RangeReplaceableCollection {
+  /// Returns a `RangeReplaceableCollection` whose length is expanded
+  /// to `paddedCount` by repeating the padding `element` at the start.
```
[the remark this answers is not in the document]
Sameesunkaria: Replaced `RangeReplacableCollection` with "new collection of the same type".

timvermeulen (approved):

timvermeulen:
@swift-ci Please test

Sameesunkaria:
Looks like I forgot to update the description in Readme.

Sources/Algorithms/Pad.swift
```diff
@@ -0,0 +1,156 @@
...
+  /// collection's length is greater than or equal to the `paddedCount`, where
+  /// _n_ is the length of the collection. For a `RandomAccessCollection`
+  /// when the collection's length is greater than or equal to the `paddedCount`,
+  /// the complexity is reduced to O(_1_).
+  @inlinable
+  public mutating func padEnd(
+    with element: Element,
+    toCount paddedCount: Int
+  ) {
+    self = paddingEnd(with: element, toCount: paddedCount)
```
LucianoPAlmeida, line 154: This is more a question but since this is mutating, would be more efficient just to `self.append(contentsOf: repeatElement(element, count: paddedCount - count))`. I'm just asking because it seems like call the no mutating version and assign to self will allocate a whole new buffer and re-append all elements and I'm not sure if this is really necessary. I didn't read all the discussion collapsed, so sorry if this was already pointed out =]

LucianoPAlmeida: One could make this case for `padStart` as well, but correct me if I'm wrong, insert `n` at begging adds more complexity to most of the collections since all the elements had to be slided `n` positions.

timvermeulen:
@swift-ci Please test

Sources/Algorithms/Pad.swift
Sameesunkaria: I decided to do some performance tests. The tests had to capture both the cases where `capacity <= paddedCount` and `capacity > paddedCount`.

To achieve this, I used the following code to perform the benchmark:

```swift
benchmark.add(
  title: "padStart + 10",
  input: [Int].self
) { input in
  return { timer in
    var inp = input
    inp.reserveCapacity(inp.count * 2)
    timer.measure {
      inp.padStart(with: 0, toCount: inp.count + 10)
    }
  }
}

benchmark.add(
  title: "padStart * 3",
  input: [Int].self
) { input in
  return { timer in
    var inp = input
    inp.reserveCapacity(inp.count * 2)
    timer.measure {
      inp.padStart(with: 0, toCount: inp.count * 3)
    }
  }
}
```

In one case I add a constant amount of padding, and in the other case I am padding the array to 3 times its original length.

The int array input provided by the `swift-collections-benchmark` library always has the same capacity as its length. To aid testing, I reserve 2x the initial capacity (simulating geometric growth). Now, for the case of constant padding, as the size of the input array increases, we would not have to allocate a new buffer. For 3x padding case, we will always allocate a new buffer.

Here are the results:

### padEnd

chart padend2

implementation of `padEnd2`:

```swift
  mutating func padEnd2(
    with element: Element,
    toCount paddedCount: Int
  ) {
    let padElementCount = paddedCount - count

    // Early exit if no padding is required.
    guard padElementCount > 0 else { return }

    self.append(contentsOf: repeatElement(element, count: padElementCount))
  }
```

**Summary:** Results are unsurprising, both algorithms performs the same when a new buffer is allocated, however `padEnd2` has a constant time (as padding length is fixed in this test case; the try complexity should be dependent on the padding length; padding length = paddedCount - collection.count) when no new buffer allocation is required.

This is a good candidate for optimization.

### padStart

chart padstart2

implementation of `padStart2`:

```swift
  public mutating func padStart2(
    with element: Element,
    toCount paddedCount: Int
  ) {
    let padElementCount = paddedCount - count

    // Early exit if no padding is required.
    guard padElementCount > 0 else { return }

    insert(contentsOf: repeatElement(element, count: padElementCount), at: startIndex)
  }
```

**Summary:** Surprisingly `padStart2` performs better in the case where no buffer allocation is required. The overall complexity of both algorithms should be the same. In the case where a new buffer allocation is required, both algorithms perform similarly.

Based on these results, I guess we can change out the implementation for `padStart` as well. Though I am not sure if these results will hold for all types of `RangeReplaceableCollection`s. What do you think?

<hr>

I believe due to how I was performing the performance tests, the results has more deviation than normal. However, after running the tests several times I always see a similar trend.

LucianoPAlmeida: Those benchmarks were nice, to me `padEnd` is definitely worth it, but maybe the guys from the swift team have some other considerations? 

> Based on these results, I am not sure if we have a clear answer here. What do you think?

For `padStart` this is somewhat expected as well, and I agree that is not a clear answer. I think  is good that for small padElementCount there is an improvement ... but if is worth it or not because of the difference on `* 3`. A hybrid version where choosing between one or another based on `padElementCount` would be an option? Not really sure.

timvermeulen: Very interesting benchmarks, especially the `padStart` one. While the time complexity of reusing the current buffer and allocating a new buffer is the same, I wonder if in the latter case an overwhelming portion of the total runtime goes towards the allocation itself which `insert(contentsOf:at:)` can (sometimes) avoid.

Either way, both of these changes look good to me. The slight slowdown in certain cases doesn't concern me too much, because
* it only appears starting at a few hundred elements which I don't think is a typical use case for these methods, and
* there's no reason I can think of why inserting elements at the start should be slower than putting it into a new buffer manually — this looks more like an optimization `insert(contentsOf:at:)` could make than something `padStart` should need to worry about.

Sameesunkaria: >A hybrid version where choosing between one or another based on padElementCount would be an option?

The only way to allow a hybrid approach would be to know the capacity before hand and switch based on if an allocation is required. A `RangeReplaceableCollection` allows you reserve capacity but asking for a collection's capacity is not part of the protocol (presumably not all `RangeReplaceableCollection`s have a notion of capacity, the default implementation of reserveCapacity is empty).

Sameesunkaria: >I wonder if in the latter case an overwhelming portion of the total runtime goes towards the allocation itself which insert(contentsOf:at:) can (sometimes) avoid.

Looks like it. In the original implementation, an entire new buffer is allocated. `insert(contentsOf:at:)` should be able to work around it by shifting all the elements in the buffer, as in the case of `padStartX + 10` we know that buffer has enough capacity in all cases for element count > 0 (where `padStart2 + 10` starts diverging).

Sameesunkaria: I guess I will change out implementations for both the mutating variants. (The complexity description for padEnd is gonna be interesting to write as there are 4 different trends based on various conditions 🤷)

timvermeulen: > I guess I will change out implementations for both the mutating variants. (The complexity description for padEnd is gonna be interesting to write as there are 4 different trends based on various conditions 🤷)

Usually costs associated with allocating new memory in case the capacity is insufficient aren't mentioned very explicitly, so don't worry about that too much.

Sameesunkaria: For padEnd I have the following 4 cases

General case:
- O(m); count < paddedCount
- O(n); count >= paddedCount

For RandomAccessCollections:
- O(m-n); count < paddedCount
- O(1); count >= paddedCount

m = paddedCount
n = collection.count

timvermeulen: Ah yeah. You could possibly condense this by defining `m` to be the number of elements that gets added instead: that would make it `O(m)` for random-access collections and `O(n + m)` otherwise, and that also makes it a bit more intuitive what the extra `n` term comes from.

lorentey:
>> I can't explain why the results change when another case is added. Could it be related to some additional compiler optimizations or speculative execution?
>
> I'm not completely sure, perhaps @lorentey could comment on this. But let's go with option 3, with the slight modification of computing count only once as it can be an expensive operation.

Benchmarking is tricky!

My (unverified, largely uninformed) hypothesis is that adding new code can change the layout of the generated binary in a way that affects the effectiveness of the instruction cache in unrelated parts. (The swift benchmark suite (and Utils/run-benchmarks.sh in Swift Collections) uses `-align-module-to-page-size` to try mitigating this, but it operates on modules, so it's unlikely to have much effect in this case. Moving each task to a separate module wouldn't be very practical.)

(Take this with a grain of salt though -- I have no idea how instruction caches work, and I should probably read up on it.) :-)

Sources/Algorithms/Pad.swift
```diff
@@ -0,0 +1,163 @@
...
+  ///     myString.padStart(with: " ", toCount: 20)
+  ///     myString // "       Hello, World!"
+  ///
+  /// - Parameters:
+  ///   - element: The element which is repeated to expand the `Collection`.
+  ///   - paddedCount: The length of the padded `Collection`.
+  ///
+  /// - Complexity: O(_m_) when the collection's length is less than the
+  /// `paddedCount`, where _m_ is equal to the `paddedCount`. O(_1_) for
+  /// a `RandomAccessCollection`; otherwise O(_n_) when the collection's length
```
LucianoPAlmeida, line 72: That wouldn't be `O(m)` given the complexity of `insert(contentsOf:at:)` is `O(m + n)` where n is length of the original collection and m is the length of newElements(which in this case is paddedCount) ?

xedin: Even if `m` was equal to `n` it would be `O(m + m)` -> `O(2 * m)` -> `O(m)` because constants could be dropped.

LucianoPAlmeida: Ah yeah, you right. It seems like I read it wrong, the statement is correct then =]

LucianoPAlmeida:
> @LucianoPAlmeida @timvermeulen I have updated the implementation for `padStart` and `padEnd`.

Looks great @Sameesunkaria :)

LucianoPAlmeida:
@swift-ci Please test

Sources/Algorithms/Pad.swift
Sameesunkaria: If we are talking about the case where the `collection.count` >= `paddedCount`, the complexity is O(_n_) because we simply need to compute the `count` and exit early as no padding is required.

LucianoPAlmeida: yeah, the statement is correct is just that I read it wrong the conditions 😅

Sameesunkaria:
@timvermeulen What are the next steps here?

Sameesunkaria:
Hey folks! Just checking up. How can I get this PR moving along?

timvermeulen:
@Sameesunkaria Sorry for the delays, we wanted to get 1.0 out the door first. Looks like this branch needs to be updated, and then let's land this!

Sameesunkaria:
I have merged in main so its easier to see the diff. But finally I would like to squash everything into a commit if allowed.

LucianoPAlmeida:
@swift-ci Please test

xwu (commented):
I think this needs either some rewording of the documentation or some reworking of the actual code:

The overwhelmingly most common use case contemplated here is padding of strings. The method claims to add padding "until" the string reaches a given length (side note: probably best to stick with consistent Swift terminology here and call it character count).

_However_, uniquely in the case of Swift strings and thanks to the magic of Unicode, combining two strings of counts `m` and `n` doesn't guarantee you a string of count `m + n` because Unicode grapheme clusters can combine. If this were some minor use case, then it might be understandable, but that's not the case here. _This method absolutely does not pad a string "until" it reaches a certain count_.

There are, as I see it, three possible solutions here:
- Document what the method actually does and, in a warning, that it does _not_ guarantee a string padded to a certain number of characters, as well as alternatives (e.g., pad to a certain number of UTF-8 or UTF-16 elements). Users may not see or understand the warning though.
- Add a check at the end of each method that the desired count is reached, and then adding more padding if it's not. This adds an extra and potentially costly call to `count` for non-string types. It's not clear to me, however, that there's not some case where combining two strings could end up with _more_ characters than expected, and it's not clear to me how best to recover from that scenario when _too much_ padding has been added.
- Add implementations specifically for `StringProtocol` types that actually do add padding "until" a count is reached, with a slow path when there are potentially combining characters involved. This seems like it'd be the hairiest and may require special Unicode know-how.

In both the latter two cases, consideration needs to be given as to the desired behavior if the supplied padding is a single character that combines entirely with the next or previous character such that the desired count can never be reached, because adding an infinite number of the combining character still leaves the string with the same character count.

timvermeulen:
@xwu Thanks for bringing this up! You're absolutely right that these methods don't necessarily do what they say they do when strings are involved, and we should document this given that these will primarily be used with strings.

That said, passing a combining character to this method, or starting out with a string that starts or ends with a combining character that the padding character combines with, is unusual enough that I don't think it's necessary to change the behavior or signature of any of these methods. Checking the `count` of the collection at the end of the operation invokes a fairly severe performance penalty that is unnecessary for just about all use cases, and it's unclear what the best way to recover from a count mismatch would be.

xwu:
Yeah, I guess I'm just voicing that, personally, I feel leery of providing 'pad' methods that are clearly geared towards strings (even with the name) that don't--and as you say, can't practically--do what they say they do.

Perhaps if they were not named and documented in a way that invites such use it would be less glaring of an issue. But, recalling that the objective of this project (as stated) is to vend canonical implementations that "improve the correctness and performance" of user code...

Sameesunkaria:
For what it's worth; using the existing implementation, if I try to padEnd `Cafe` to the length of `10` using an acute accent (`\u{0301}`), the final count of the string remains unchanged at `4`. 

All the combining characters I tried showed the same behavior. So, I can't really see a way to ever "pad" a string to the desired length in Swift when the padding character is free to be any Unicode character.

xwu:
@Sameesunkaria For what it's worth, I'm not overly excited about the case where you're using an isolated combining character like the acute accent, which is (I believe) considered a "degenerate" case that explicitly shouldn't have special handling.

But not all cases where strings don't pad the way you'd expect involve such "degenerate" cases. For example (as we've just been discussing on the forums about character classes), regional indicators are standalone grapheme clusters because they're classified as symbols. So `"🇧"` is a well-formed Unicode string with character count 1. However, if you pad a string with it, you're only going to get half as much padding as you'd expect, because two of these back-to-back encode the flag for Barbados: `"🇧🇧"`.

(Ironically, I'm on a Windows machine at the moment, and there's no emoji symbol for that flag in the system font, so I see two B's but the caret correctly skips them both when I advance with the right arrow.)

ls-samar-sunkaria:
>This cannot be determined by induction because some characters will combine only once or a finite number of times and others infinitely.

Based on this comment, if for any well-formed string we cannot generate a string of the desired length with the specified well-formed padding character, then would it even be possible to write a padding function in Swift?

If not, we may consider renaming the method to better describe its actual behavior.

Sameesunkaria:
> This cannot be determined by induction because some characters will combine only once or a finite number of times and others infinitely.

Based on this comment, if for any well-formed string we cannot generate a string of the desired length with the specified well-formed padding character, then would it even be possible to write a padding function in Swift?

If not, we may consider renaming the method to better describe its actual behavior.

Sameesunkaria:
So, folks where do we go from here?

Based on the discussions we had above, I see three possible options. I have gathered some comments here.

1. Keep the same implementation (possibly changing the name) and document the behavior for strings. However, I agree with @xwu that we must be careful while providing such APIs.
>[...] I feel leery of providing 'pad' methods that are clearly geared towards strings (even with the name) that don't--and as you say, can't practically--do what they say they do. Perhaps if they were not named and documented in a way that invites such use it would be less glaring of an issue, and then it's not the behavior that needs modifying.
&nbsp;
But, recalling that the objective of this project (as stated) is to vend canonical implementations that " improve the correctness and performance](https://swift.org/blog/swift-algorithms/)" of user code... In no small part that comes from the canonical implementation handling the "unusual enough" corner cases [...]
&nbsp;
https://github.com/apple/swift-algorithms/pull/157#issuecomment-948062591

2. Find a way to add additional elements if the desired count is not reached (may be a `StringProtocol` specific implementation). This approach also has its caveats; the most prominent one being that _we don't know if it's possible to implement_. Performance being another issue. Finally if we decide to go for a _best-effort_ approach where we try to get close to the final count, what would qualify as best-effort?
>[...] consideration needs to be given as to the desired behavior if the supplied padding is a single character that combines entirely with the next or previous character such that the desired count can never be reached because adding an infinite number of the combining character still leaves the string with the same character count. This **cannot be determined by induction because some characters will combine only once or a finite number of times and others infinitely**.
&nbsp;
https://github.com/apple/swift-algorithms/pull/157#pullrequestreview-752167700

> [...] Checking the count of the collection at the end of the operation invokes a fairly severe performance penalty that is unnecessary for just about all use cases, and it's unclear what the best way to recover from a count mismatch would be.
&nbsp;
https://github.com/apple/swift-algorithms/pull/157#issuecomment-947610660

3. Acknowledge defeat. A padding API is often exclusively used with strings. IMO dismissing the behavior of this API for strings goes agains the objective of this project ([also pointed out by @xwu](https://github.com/apple/swift-algorithms/pull/157#issuecomment-948062591)). Since Swift `String`'s conformance to `Collection` is implemented on the character view, we may never be able to provide a correct string padding API.

What do you think about it @timvermeulen @xwu?

xwu:
I don't make the calls here, but personally I think if we can find a non-leading name for this feature (in the legal sense of "leading" as in suggestive of something--here, suggestive of being appropriate for strings), and so definitely not using the term "pad," and we document _very_ well and point out how _well-formed_ strings (not just you examples using degenerate cases) can have unusual behavior, then it could work out.

A question to consider is, how often to folks want to pad _non_-strings? If the answer is comparatively much more rarely than for strings, then the alternative of vending a proper string-specific implementation--and I'd imagine in the standard library rather than here (since it would require investigating whether there are grapheme-breaking APIs we need so that this can be done without iterative guess-and-check and potential infinite loops) would be the way to go.

PR #165, opened 2021-08-27
[Rotate] Modify API of rotate(subrange:) to take an arbitrary range expression

LucianoPAlmeida:
Just throwing the idea here, maybe would be nice to be able to call this API with any kind of range. 
And since is not a source breaking change, I thought it was worth bring to discussion. 

cc @natecook1000 

### Checklist

### DIFF
--- a/Sources/Algorithms/Rotate.swift
+++ b/Sources/Algorithms/Rotate.swift
@@ -140,9 +140,9 @@ extension MutableCollection {
     subrange: Range,
     toStartAt newStart: Index
   ) -> Index where Range.Bound == Index {
-    let cSubrange = subrange.relative(to: self)
-    var m = newStart, s = cSubrange.lowerBound
-    let e = cSubrange.upperBound
+    let relativeSubrange = subrange.relative(to: self)
+    var m = newStart, s = relativeSubrange.lowerBound
+    let e = relativeSubrange.upperBound
     
     // Handle the trivial cases
     if s == m { return e }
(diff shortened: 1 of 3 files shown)


mdznr:
I think this is a great change. Since you could subscript a collection this way, it makes sense for `rotate` to also let you reference a subrange this way.

Sources/Algorithms/Rotate.swift
```diff
@@ -132,15 +132,17 @@ extension MutableCollection {
...
-  public mutating func rotate(
-    subrange: Range<Index>,
+  public mutating func rotate<Range: RangeExpression>(
+    subrange: Range,
     toStartAt newStart: Index
-  ) -> Index {
-    var m = newStart, s = subrange.lowerBound
-    let e = subrange.upperBound
+  ) -> Index where Range.Bound == Index {
+    let cSubrange = subrange.relative(to: self)
```
xwu, line 143: Question: what does the "c" in `cSubrange` mean?

LucianoPAlmeida: I really don't remember sorry, but changed to `rSubrange` for `relative` as it makes more sense I guess.

xwu: Perhaps `relativeSubrange` will help everyone understand?

LucianoPAlmeida: Just fixed

PR #167, opened 2021-09-07
Prepare for Swift Algorithms 1.0

timvermeulen:


timvermeulen:
@swift-ci Please test

timvermeulen:
@swift-ci Please test

CHANGELOG.md
```diff
@@ -4,18 +4,81 @@
...
+  LazySplitCollection -> SplitCollection
+  Permutations        -> PermutationsSequence
+  UniquePermutations  -> UniquePermutationsSequence
+  Product2            -> Product2Sequence
+  ExclusiveReductions -> ExclusiveReductionsSequence
+  InclusiveReductions -> InclusiveReductionsSequence
+  StrideSequence      -> StridingSequence
+  StrideCollection    -> StridingCollection
+  Uniqued             -> UniquedSequence
+  Windows             -> WindowsCollection
```
karwa, line 46: IMO, "WindowedCollection" would be better aligned with the other type names (e.g. IndexedCollection, rather than IndexesCollection)

timvermeulen: I don't necessarily disagree, but we've arrived at these names by taking the base name of the method that produces them. In this case those are `windows(ofCount:)` and `indexed()`. The renaming of `ChunkedByCount` to `ChunksOfCountCollection` is another such example.

karwa: 🤷‍♂️ All of the other names seem fine, but this one just seems a bit off. Given that this is the 1.0 release, I just think it’s worth reconsidering is all. WindowsOfCountCollection, perhaps? Still a bit verbose for my liking, but better than WindowsCollection (which reads like something to do with Microsoft’s OS).

timvermeulen: @karwa Thanks for suggesting `WindowsOfCountCollection`! It indeed clears up any ambiguity and is still consistent with the other type names in this package, so that's the one we're going with.

timvermeulen:
@swift-ci Please test

timvermeulen:
@swift-ci Please test

PR #174, opened 2021-11-28
Add @discardableResult attribute to stablePartition

adincebic:
- [x] I've completed this task
- [ ] This task isn't completed

Since stablePartition is used frequently in contexts where it's return result isn't needed, I added @discardableResult attribute to both public methods for stablePartition to avoid Xcode's warning about unused result of call to stablePartition. 

### Checklist

### DIFF
--- a/Sources/Algorithms/Partition.swift
+++ b/Sources/Algorithms/Partition.swift
@@ -58,6 +58,7 @@ extension MutableCollection {
   ///
   /// - Complexity: O(*n* log *n*), where *n* is the length of this collection.
   @inlinable
+  @discardableResult
   public mutating func stablePartition(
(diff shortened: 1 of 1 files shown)


adincebic:
@swift-ci Please smoke test

glessard:
@swift-ci please test

glessard:
Why add `@discardableResult` on `stablePartition` but not `partition`?

natecook1000:
Thanks for your contribution, @adincebic! I don't think there are _frequent_ cases where you want to partition a collection but don't care about where the partition is — partitioning is generally used as part of a larger algorithm, and further processing depends on the position of the partition. Feel free to start a thread on the Swift forums if you'd like to discuss further!

PR #175, opened 2021-11-28
[Rotate] Non-mutating variants for rotating APIs

LucianoPAlmeida:
This is an implementation of non-mutating variant for `rotate(toStartAt:)` and `rotate(subrange:toStartAt:)` 
that are `rotated(toStartAt:)` and `rotated(subrange:toStartAt:)` respectively that results in a `RotatedCollection` view.
The useful aspects of such API is that when preserving the original collection, the mutating version requires a copy. Also, when base conforms to `RandomAccessCollection` calling `rotated` is O(1) time and elements at each position are also computed in constant time. 

Couple of other points to note:
* `Rotate.md` documentation was updated with new variants. Not sure if was worth a new file both for implementation or docs. So opted to keep everything together. 
* Tests to ensure that both mutating and non-mutating variants produce the same result.
* There is also a point worth discussing together with this, that is #165, if that is a useful API modification and if both PRs are useful additions we can land together. 

Let me know what you think @natecook1000 @timvermeulen @lorentey  

### Checklist

### DIFF
--- a/Guides/Rotate.md
+++ b/Guides/Rotate.md
@@ -24,6 +24,18 @@ numbers.rotate(subrange: 3..<6, toStartAt: 4)
 // numbers = [20, 30, 10, 50, 60, 40]
 ```
 
+We also provide non-mutating variants of `rotate(toStartAt:)` and `rotate(subrange:toStartAt:)` 
+that are `rotated(toStartAt:)` and `rotated(subrange:toStartAt:)` respectively and results in 
+a `RotatedCollection` view that presents the elements of a `base` collection rotated. 
+
+```swift
+var numbers = [10, 20, 30, 40, 50, 60]
+let first = numbers.rotated(subrange: 0..<3, toStartAt: 1)
+// first = [20, 30, 10, 40, 50, 60]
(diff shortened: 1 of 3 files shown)


LucianoPAlmeida:
A quick bench about difference in perf for a single all elements iteration just out of curiosity 
```swift
var benchmark = Benchmark(title: "Rotated Algorithm")
let inputGenerator: (Int) -> [Int] = { size in
  return (0..<size).map { $0 }
}
benchmark.registerInputGenerator(inputGenerator)

benchmark.add(
  title: "In place rotation 25%",
  input: [Int].self
) { input in
  var mInput = input
  let count = mInput.count / 4 // 25% rotation
  return { timer in
    mInput.rotate(toStartAt: count)
    for value in mInput {
      blackHole(value)
    }
  }
}

benchmark.add(
  title: "Rotated View 25%",
  input: [Int].self
) { input in
  let count = input.count / 4 // 25% rotation
  return { timer in
    for value in input.rotated(toStartAt: count) {
      blackHole(value)
    }
  }
}

benchmark.main()
```
chart

PR #177, opened 2021-12-21
`XCTAssertUnorderedEqualSequences`: Improve algorithmic complexity when elements are `Hashable` (Resolves #176)

### LINKED ISSUE #176: `XCTAssertUnorderedEqualSequences`: Improve algorithmic complexity when elements are `Hashable`
In the (only) implementation of `XCTAssertUnorderedEqualSequences` (where elements are `Equatable`), it converts the sequence into an `Array`. Then it iterates through the second sequence, calling `firstIndex(of:)` on the first array (O(*n*)), which totals to O(*n*<sup>2</sup>). If a `Set` were used (when the elements are `Hashable`), then getting the index of an element is O(1), which would make the total algorithmic complexity O(*n*).

#### Expected behavior
`XCTAssertUnorderedEqualSequences ` is O(*n*) when `Element: Hashable`

#### Actual behavior
`XCTAssertUnorderedEqualSequences ` is O(*n*<sup>2</sup>) when `Element: Hashable`

mdznr:
Resolves https://github.com/apple/swift-algorithms/issues/176

### Checklist

### DIFF
--- a/Tests/SwiftAlgorithmsTests/TestUtilities.swift
+++ b/Tests/SwiftAlgorithmsTests/TestUtilities.swift
@@ -86,6 +86,8 @@ func XCTAssertEqualSequences<S1: Sequence, S2: Sequence>(
 
 /// Asserts two sequences contain exactly the same elements but not necessarily
 /// in the same order.
+/// - Complexity: O(*n* * *m*) where *n* is the number of elements in the first
+/// sequence and *m* is the number of elements in the second sequence
 func XCTAssertUnorderedEqualSequences<S1: Sequence, S2: Sequence>(
   _ expression1: @autoclosure () throws -> S1,
   _ expression2: @autoclosure () throws -> S2,
--- a/Tests/SwiftAlgorithmsTests/TestUtilities.swift
+++ b/Tests/SwiftAlgorithmsTests/TestUtilities.swift
@@ -117,7 +117,52 @@ func XCTAssertUnorderedEqualSequences<S1: Sequence, S2: Sequence>(
   if !s1.isEmpty {
     fail("first sequence contains \(s1) missing from second sequence")
   }
+}
 
+/// Asserts two sequences contain exactly the same elements but not necessarily
+/// in the same order.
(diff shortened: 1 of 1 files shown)


Tests/SwiftAlgorithmsTests/TestUtilities.swift
```diff
@@ -100,15 +110,59 @@ func XCTAssertUnorderedEqualSequences<S1: Sequence, S2: Sequence>(
     s1.remove(at: idx)
```
LucianoPAlmeida, line 110: When it was implemented perf was not in mind because well it was just testing. There is maybe a small optimization that can be done here as well.
```suggestion
    // Since ordering does not matter, this is a small
    // optmization that avoid removing a index in the middle
    // of the array and having to slide all elements.
    let lastIdx = s1.index(before: s1.endIndex)
    if lastIdx != idx {
      s1.swapAt(idx, lastIdx)
    }
    s1.remove(at: lastIdx)
```
So by swapping always to back and removing there, we avoid removal in middle which my guess need to slide all the elements after one position after removal. 

Didn't test it, is just a guess, but throwing the idea here in case you want to give it a shot since you are measuring as well =]

mdznr: The logic makes sense to me. I did a quick benchmark on two random arrays of integers and there is consistent performance improvements in larger collections. This is an interesting use case for an unordered collection, like `Set` but without uniqueness or the `Hashable` requirements.

chart

mdznr: I updated the documentation and removed the need for the redundant `!=` check since that already occurs internally in [`swapAt(_:_:)`](https://github.com/apple/swift/blob/59559ca0ce668b0240bcb2a91f874922a8eba2d7/stdlib/public/core/MutableCollection.swift#L321).

LucianoPAlmeida: Awesome! Thanks for the measurements :)

PR #182, opened 2022-02-03
Correct Chunked documentation

mdznr:
- Corrected documentation in Chunked.md regarding the algorithmic complexity of the Chunked functions
- Added parameter documentation for `belongInSameGroup` and `projection`

### Checklist

### DIFF
--- a/Guides/Chunked.md
+++ b/Guides/Chunked.md
@@ -96,7 +96,8 @@ collection conforms.
 
 ### Complexity
 
-The eager methods are O(_n_), the lazy methods are O(_1_).
+The eager methods are O(_n_) where _n_ is the number of elements in the
+collection. The lazy methods are O(_n_) because the start index is pre-computed.
 
 ### Naming
 
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -218,7 +218,7 @@ extension LazySequenceProtocol where Self: Collection, Elements: Collection {
   }
   
   /// Returns a lazy collection of subsequences of this collection, chunked by
-  /// grouping elements that project to the same value.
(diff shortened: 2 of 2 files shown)


Guides/Chunked.md
```diff
@@ -96,7 +96,8 @@ collection conforms.
 
 ### Complexity
 
-The eager methods are O(_n_), the lazy methods are O(_1_).
+The eager methods are O(_n_) where _n_ is the number of elements in the
+collection. The lazy methods are O(_n_) because the start index is pre-computed.
```
mdznr, line 100: There definitely is a difference in complexity between the eager and lazy functions, but I’m finding it difficult to describe succinctly.

The eager functions are O(_n_) where _n_ is the number of elements in the collection.

The lazy functions on the order of O(_n_) since the start index could require iterating to the very end of the collection. Each iteration of the receiving lazy sequence is also on the order of O(_n_), but total _n_ where _n_ is the number of elements in the collection—not _n_<sup>2</sup>.

Sources/Algorithms/Chunked.swift
```diff
@@ -218,7 +221,11 @@ extension LazySequenceProtocol where Self: Collection, Elements: Collection {
   }
   
   /// Returns a lazy collection of subsequences of this collection, chunked by
-  /// grouping elements that project to the same value.
+  /// grouping elements that project to the equal values.
```
LucianoPAlmeida, line 224: `to the equal values.` Seems not to read correctly, would that be `to equal values`?

mdznr:
```suggestion
  /// grouping elements that project to equal values.
```

natecook1000:
@swift-ci Please test

PR #185, opened 2022-03-27
Add `recursiveMap(_:)`

SusanDoggie:
add `recursiveMap()` method.

### Checklist


CTMacUser:
Is a "recursive mapped sequence" a term of art for a particular structure?  I've never heard of it before, and the documentation so far assumes you pretty much already know.  A quick web search found something related using the Python language; is the code here supposed to be a Swift adaptation of the Python code?

The concept seems to be a way to do tree traversal without that type having to provide its own depth- and breath-first iterators; you just have to supply a function to map an element to a `Sequence` of its children, and this function takes care of the rest.

SusanDoggie:
> Is a "recursive mapped sequence" a term of art for a particular structure? I've never heard of it before, and the documentation so far assumes you pretty much already know. A quick web search found something related using the Python language; is the code here supposed to be a Swift adaptation of the Python code?
> 
> The concept seems to be a way to do tree traversal without that type having to provide its own depth- and breath-first iterators; you just have to supply a function to map an element to a `Sequence` of its children, and this function takes care of the rest.

This method is simplified recursive CTE, which can map a hierarchical table to a linear sequence.

I don’t known what’s the name of this operation. So I call it recursive map.

Sources/Algorithms/RecursiveMap.swift
```diff
@@ -0,0 +1,187 @@
...
+    /// // 6
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - option: Traversal option. This option affects the element order of the output sequence. default depth-first.
+    ///   - transform: A closure that map the element to new sequence.
+    /// - Returns: A sequence of the original sequence followed by recursive mapped sequence.
+    ///
+    /// - Complexity: O(1)
+    @inlinable
```
ssouzawallace, line 48: @SusanDoggie 
I'm kinda skeptic this func is actually O(1) for **time-complexity**. Or any recursive function. Unles I'm missing a rare 0.00001% case. Are you sure about that L47 statement?

Could you demonstrate if possible?

Looks like it is at least O(n). But what is n? Is n self.count? Is n the whole tree size i.e. all nodes from self.

Also *VERY IMPORTANT**! It doesn't look like your Node struct example is conforming with Sequence or RecursiveMapSequence at all.

Other than that. Interesting use case you put there and awesome work with putting this so much ready for official documentation.

ssouzawallace (changes requested):
Without further proof of time and space complexity. This PR IMO is unmeargeable.
Even though it could be a good addition. I firstly thought that this would apply to any Sequence. However. Looks like this is a special case of Colletion and maybe not Sequence.
Having to conform to RecursiveMapSequence seems like too much trouble for a rare case **in this protocol**.
But as I said. Maybe it fits better with Collection or similar.

-/+ 1 - The Node struct example is not properly written for official docs.

ssouzawallace:
@SusanDoggie  Also. Tip. If you plan on keeping this PR up-to-date. I suggest squash commits, make them atomic commits, all properly signed.

PR #186, opened 2022-04-20
Update Suffix.swift

mattyoung:
Change `endOfPrefix(while:)` and `startOfSuffix(while:)` from internal to public

These two functions are generally useful so they should be public instead of internal
See https://forums.swift.org/t/should-there-be-bidirectionalcollection-droplast-while/48059/14
and https://forums.swift.org/t/endofprefix-while-and-startofsuffix-while-should-be-public-instead-of-internal/56855

### Checklist

### DIFF
--- a/Sources/Algorithms/Suffix.swift
+++ b/Sources/Algorithms/Suffix.swift
@@ -46,7 +46,7 @@ extension Collection {
   ///
   /// - Complexity: O(*n*), where *n* is the length of the collection.
   @inlinable
-  internal func endOfPrefix(
+  public func endOfPrefix(
     while predicate: (Element) throws -> Bool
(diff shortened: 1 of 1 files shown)


lucianopa-msft:
Since we are making those public, couple of points are think are worth mention:
* We probably need to add this to Changelog.md
* We also may need a `Guide` the same way we have for all other algorithms here. See [Guides](https://github.com/apple/swift-algorithms/tree/main/Guides) for reference.
* Add entry to the Readme.md
* Verify if we have tests for those methods? I believe we test that indirectly through the public methods that use them internally, but not sure we have tests for them specifically (probably have, but maybe it is worth check)

natecook1000:
Landing this change in #211.

PR #197, opened 2023-06-21
Add `grouped(by:)` and `keyed(by:)`

amomchilov:
### Description

Pitch: https://forums.swift.org/t/pitch-sequence-grouped-by-and-keyed-by/65511
Swift Evolution proposal: https://github.com/apple/swift-evolution/pull/2078/files?short_path=2767014

### Detailed Design
Include any additional information about the design here. At minimum, show any new API:

```swift
extension Sequence {
  func grouped<GroupKey>(
    by keyForValue: (Element) throws -> GroupKey
  ) rethrows -> [GroupKey: [Element]]
  
  func keyed<Key>(
    by keyForValue: (Element) throws -> Key,
    uniquingKeysWith combine: ((Key, Element, Element) throws -> Element)? = nil
  ) rethrows -> [Key: Element]
}
```

### Documentation Plan

I've updated new `Grouping.md` and `Keyed.md` guides, and updated README to point to thme.

### Test Plan

I wrote tests loosely based on the Standard Library's [tests for `Dictionary.init(grouping:by:)`](https://github.com/apple/swift/blob/4d1d8a9de5ebc132a17aee9fc267461facf89bf8/validation-test/stdlib/Dictionary.swift#L1974-L1988). The Standard Library has a magical `expectCrashLater()` which lets you test fatalErrors and such. Without it, I'm not sure how to test properly test the `fatalError()` this PR introduces to `keyed(by:)`.

### Source Impact

Purely additive.

### Checklist


Guides/Grouped.md
```diff
@@ -0,0 +1,71 @@
...
+
+```swift
+let fruits = ["Apricot", "Banana", "Apple", "Cherry", "Avocado", "Coconut"]
+let fruitsByLetter = fruits.grouped(by: { $0.first! })
+// Results in:
+[
+    "B": ["Banana"],
+    "A": ["Apricot", "Apple", "Avocado"],
+    "C": ["Cherry", "Coconut"],
+]
```
amomchilov, line 19: This is a different style to how other examples show their output, but I think it's better to be multi-line and syntax highlighted in this case. Compare:

```swift
// fruitsByLetter == ["B": ["Banana"],  "A": ["Apricot", "Apple", "Avocado"], "C": ["Cherry", "Coconut"]]
```

Guides/Grouped.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+If you wish to achieve a similar effect but for single values (instead of Arrays of grouped values), see [`keyed(by:)`](Keyed.md).
```
amomchilov, line 22: These two APIs are similar enough that I think it's worth cross-linking between the two as a convenience for readers.

Guides/Grouped.md
```diff
@@ -0,0 +1,71 @@
...
+#### Customization points
+
+Java and C# are interesting in that they provide multiple overloads with several points of customization:
+
+1. Changing the type of the groups.
+    1. E.g. the groups can be Sets instead of Arrays.
+    1. Akin to calling `.transformValues { group in Set(group) }` on the resultant dictionary, but avoiding the intermediate allocation of Arrays of each group.
+2. Picking which elements end up in the groupings.
+    1. The default is the elements of the input sequence, but can be changed.
+    2. Akin to calling `.transformValues { group in group.map(someTransform) }` on the resultant dictionary, but avoiding the intermediate allocation of Arrays of each group.
```
amomchilov, line 66: Side thought: If `grouped(by:)` was instead implemented to be lazy, the `keyed(by: \.key)` could be replaced with `grouped(by: \.key).transformValues(\.last)`. It's still wordy though, so I think `keyed(by:)` is keeping as-is.

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,70 @@
...
+    if combine != nil {
+      // We have a `combine` closure. Use it to resolve duplicate keys.
+
+      for element in self {
+        let key = try keyForValue(element)
+        
+        if let oldValue = result.updateValue(element, forKey: key) {
+          // Can't use a conditional binding to unwrap this, because the newly bound variable
+          // doesn't play nice with the `rethrows` system.
+          let valueToKeep = try combine!(key, oldValue, element)
```
amomchilov, line 46: If you used `if let combine` above, `combine` will refer to a new variable and not the parameter, so the `rethrows` support blows up:

```
// Call can throw, but the error is not handled; a function declared 'rethrows' may only throw if its parameter does
```

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,70 @@
...
+          
+          // This causes a second look-up for the same key. The standard library can avoid that
+          // by calling `mutatingFind` to get access to the bucket where the value will end up,
+          // and updating in place.
+          // Swift Algorithms doesn't have access to that API, so we make due.
+          // When this gets merged into the standard library, we should optimize this.
+          result[key] = valueToKeep
```
amomchilov, line 53: This is what I'm referring to: https://github.com/apple/swift/blob/7c3917c0d38a6f29e68c13a7c1e89d74e7ccb720/stdlib/public/core/NativeDictionary.swift#L755-L776

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+      }
+    } else {
+      // There's no `combine` closure. Duplicate keys are disallowed.
+
+      for element in self {
+        let key = try keyForValue(element)
```
amomchilov, line 60: I repeat this loop boilerplate, so I don't have to check `combine` on each iteration. 

TODO: Is it worth it? Can the compiler identify that `if combine != nil` is invariant, and hoist it out of the loop?

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,70 @@
...
+  /// - Parameters:
+  ///   - keyForValue:  A closure that returns a key for each element in
+  ///     `self`.
+  ///   - combine: A closure that is called with the values for any duplicate
+  ///     keys that are encountered. The closure returns the desired value for
+  ///     the final dictionary.
+  @inlinable
+  public func keyed<Key>(
+    by keyForValue: (Element) throws -> Key,
+    uniquingKeysWith combine: ((Key, Element, Element) throws -> Element)? = nil
```
amomchilov, line 33: Is `uniquingKeysWith` still the right naming? After all, it's not the keys being uniqued, it's the values.

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+      }
+    } else {
+      // There's no `combine` closure. Duplicate keys are disallowed.
+
+      for element in self {
+        let key = try keyForValue(element)
+
+        guard result.updateValue(element, forKey: key) == nil else {
+          fatalError("Duplicate values for key: '\(key)'")
```
amomchilov, line 63: Ideally, I would have just done something like:

```swift
let combine = combine ?? { key, _, _ in fatalError("Duplicate values for key: '\(key)'") }
```

but I couldn't get that to play nice with the `rethrows` system.

Interestingly, the addition of the `key` here works out pretty nicely, as compared to what the standard library had to do:

https://github.com/apple/swift/blob/7c3917c0d38a6f29e68c13a7c1e89d74e7ccb720/stdlib/public/core/Dictionary.swift#L484-L492

Guides/Keyed.md
```diff
@@ -0,0 +1,74 @@
+# Indexed
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Indexed.swift) | 
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/IndexedTests.swift)]
```
natecook1000, line 4: Check the title and links here…

Sources/Algorithms/Keyed.swift
natecook1000: Instead of making the closure `nil`, what about providing the default behavior as a closure?

Sources/Algorithms/Keyed.swift
natecook1000: I don't think it will be quite as efficient as the stdlib version, but you can eliminate the duplicated lookup and hashing by using `index(forKey:)` and then mutating `result.values[i]` instead of using the key-based subscript.

Tests/SwiftAlgorithmsTests/KeyedTests.swift
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  func testEmpty() {
+    let d = EmptyCollection<String>().keyed(by: { $0.first! })
+    XCTAssertEqual(d.count, 0)
+  }
+
+  func testNonUniqueKeys() throws {
+    throw XCTSkip("""
+    TODO: What's the XCTest equivalent to `expectCrashLater()`?
```
natecook1000, line 34: Can't test this with XCTest, unfortunately.

natecook1000 (commented):
Looks good so far! I have two questions about the design:

1. The `keyed(by:)` and `grouped(by:)` methods seem very related — do you think it would make sense to include a method like this (naming tbd) that incorporates/enables them both? `grouped(by:)` is a specialization of this, as is a "frequencies" method if I'm thinking of it correctly (e.g. `seq.indexed(by: \.self, transform: { _ in 1 }, merge: +)`)

    ```swift
    func keyed<Key: Hashable, NewElement>(
        by: (Element) -> Key,
        transform: (Element) throws -> NewElement,
        merge: (NewElement, NewElement) throws -> NewElement
    ) rethrows -> [Key: NewElement]
    ```

2. I don't know that a crash is the right default behavior for `keyed(by:)` — for the dictionary initializer we require the `unique` parameter name so that an author is essentially asserting the right behavior. Do you think crashing there is okay without any other warning, or would a "last value wins" approach be acceptable? (That would match the behavior of most other implementations and replicate the trivial code that this method replaces.)

Sources/Algorithms/Keyed.swift
amomchilov: I tried that, but I couldn't make it work with the `rethrows` inference

Sources/Algorithms/Keyed.swift
amomchilov: I don't think that quite works. It solves the case of double-lookups for conflicting keys, but it shifts the problem to the initial assignment:

```swift
if let existingIndex = result.index(forKey: key) { // First hash
    let valueToKeep = try combine!(key, oldValue, element)
    result.values[existingIndex] = valueToKeey // No longer needs a second hash
} else {
    // The item is new and unique
    result[key] = element // Second hash... the problem moved to this (more common) case
}
```

Guides/Grouped.md
```diff
@@ -0,0 +1,71 @@
+# AdjacentPairs
```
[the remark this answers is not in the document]
amomchilov: Yep! Copied the file, forgot to update.

amomchilov:
> do you think it would make sense to include a method like this (naming tbd) that incorporates/enables them both?

For the sake of flexibility, I would like to eventually _also_ have this API, but even if it exists, I think specialized version for these two use cases (and [`tally`](https://rubydoc.info/stdlib/core/Enumerable:tally)) is worthwhile, to simplify the common case.

Guides/Keyed.md
```diff
@@ -0,0 +1,74 @@
...
+let fruitByLetter = fruits.keyed(by: { $0.first! })
+// Results in:
+[
+	"A": "Apple",
+	"B": "Banana",
+	"C": "Cherry",
+]
+```
+
+Duplicate keys will trigger a runtime error by default. To handle this, you can provide a closure which specifies which value to keep:
```
wadetregaskis, line 19: This makes the method a bit dangerous, since a naive user could fail to realise keys might collide.  Even a knowledgeable user might make the simple mistake of forgetting to provide the `uniquingKeysWith` argument (e.g. auto-complete omits it and they don't notice in the moment).

It makes it more verbose to handle 'unsafe' data - arguably the common case in today's network-centric world - since you have to implement the second closure (as merely a throw statement) in order to avoid crashing on bad input.

These problems can be addressed by having two versions of the method:

```swift
public func keyed<Key>(
    by keyForValue: (Element) throws -> Key
) throws -> [Key: Element]

public func keyed<Key>(
    by keyForValue: (Element) throws -> Key,
    uniquingKeysWith combine: (Key, Element, Element) throws -> Element
) rethrows -> [Key: Element]
```

This preserves the ability to have a non-throwing version - and thus not require use of `try` - iff the second form is used and both closures don't throw, which is the only case where it's compile-time assured that it cannot fail.

Any user can still use `try!` to mimic the other behaviour (of crashing), if they wish.  And the explicit presence of that `!` in their code is an important warning that it is potentially unsafe.

(it also has the minor benefit of not allowing you to explicitly set `uniquingKeysWith` to nil, which has ambiguous and potentially misleading intent, e.g. a reader might assume that makes it not unique keys (somehow))

Sources/Algorithms/Keyed.swift
wadetregaskis: Good point.  It does follow convention from the `Dictionary` initialiser, though.

Maybe `resolvingConflictsWith`?

Technically this is equivalent to first grouping by the keys then reducing the resulting dictionary values… so maybe `reducingWith`?  In a way that's nicer because it reminds us that you don't _have_ to return one or other of the given values - you can technically return any value of the same type, such as the sum of both values, or the largest or smallest, etc.

There's also no way, with this design, to preserve both elements but under different keys.  But I'm not sure that needs to be supported… in such cases, perhaps `group(by:)` is sufficient?

Guides/Keyed.md
amomchilov: > These problems can be addressed by having two versions of the method:

I love this idea! I'll make these changes

> (it also has the minor benefit of not allowing you to explicitly set uniquingKeysWith to nil, which has ambiguous and potentially misleading intent, e.g. a reader might assume that makes it not unique keys (somehow))

Good call!

Sources/Algorithms/Keyed.swift
amomchilov: > Maybe `resolvingConflictsWith`?

I'm digging this. "Resolving" is clear enough in context, while succinct and not overly prescriptive as to what the closure's resolution logic should be

> There's also no way, with this design, to preserve both elements but under different keys. But I'm not sure that needs to be supported… in such cases, perhaps group(by:) is sufficient?

Agree.

Guides/Grouped.md
wadetregaskis: It'd still be notably less efficient, due (mainly) to the construction of the intermediary arrays.  I assume the Swift compiler won't be clever enough to eliminate those.

Aside from that, how would lazy execution work?  Wouldn't you have to run through the complete input sequence in order to know when you have all the values for any given key?

Guides/Grouped.md
```diff
@@ -0,0 +1,68 @@
...
+
+All the surveyed languages name this operation with a variant of "grouped" or "grouping". The past tense `grouped(by:)` best fits [Swift's API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/).
+
+#### Customization points
+
+Java and C# are interesting in that they provide multiple overloads with several points of customization:
+
+1. Changing the type of the groups.
+    1. E.g. the groups can be Sets instead of Arrays.
+    1. Akin to calling `.transformValues { group in Set(group) }` on the resultant dictionary, but avoiding the intermediate allocation of Arrays of each group.
```
wadetregaskis, line 60: ### Forward

I explored all this really just for my own edification, but it might be helpful to include some summary of this in your notes here, to explain _why_ the `group(by:)` API doesn't and _cannot_ currently support this functionality.  Adding all the necessary infrastructure - whether a `Collector` protocol or whatever else - seems like a more fundamental addition to the Swift standard library that should be tackled in its own patch & pitch.

### Rabbit hole

I like the idea of supporting this, but it doesn't seem like it's possible [elegantly] in Swift today…?  Swift's standard library & Foundation don't include an equivalent to Java's `Collector` interface, for example.  And even though I'm not a big fan of that approach anyway, you fundamentally do need a way to say "any collection that can be initialised and added to", which Swift does not have.

Specifically: you can easily genericise `group(by:)` with some `ValueCollection` type, but what protocol would it have to conform to?  `Collection` specifies only the APIs for _reading_ from a _given_ collection, not for constructing one or adding to it.  `MutableCollection` (surprisingly) has the exact same limitations, _and_ feels inappropriate anyway since it forces the resulting Dictionary to have a mutable type for its values.

There is `RangeReplaceableCollection` as a half measure - it does include an init method, but only a limited subset of standard collections conform to it (notably not including `Set`).  Still, with that you can do:

```swift
extension Sequence {
    public func grouped<GroupKey, ValueCollection: RangeReplaceableCollection<Element>>(
        by keyForValue: (Element) throws -> GroupKey
    ) rethrows -> [GroupKey: ValueCollection] {
        var result = [GroupKey: ValueCollection]()

        for value in self {
            result[try keyForValue(value), default: ValueCollection()].append(value)
        }

        return result
    }
}
```

Semantically it's a hack, of course - there's nothing about the _range replacement_ functionality actually needed here, it's just used because it exposes other APIs almost as a side-effect.

And anyway, the ergonomics are poor since Swift doesn't support specifying a default value for a generic parameter, requiring boilerplate in many cases:

```swift
[1, 2, 3, 4].grouped { $0 % 2 } // ❌ Generic parameter 'ValueCollection' could not be inferred

let result: [Int: [Int]] = [1, 2, 3, 4].grouped { $0 % 2 } // Works, but awkward for many use-cases.
```

One could work around the initialisation problem by having the caller explicitly provide an initialiser, but you still need a protocol that all collections support which specifies some kind of `append` method, and that doesn't currently exist.

It _is_ possible to support arbitrary collection types, but only by having the caller provide a reducer explicitly in order to work around the aforementioned limitations, e.g.:

```swift
extension Sequence {
    public func grouped<GroupKey, ValueCollection: Collection>(
        by keyForValue: (Element) throws -> GroupKey,
        collectionInitialiser: () -> ValueCollection = Array<Element>.init,
        collectionReducer: (inout ValueCollection, Element) -> ValueCollection
    ) rethrows -> [GroupKey: ValueCollection] where ValueCollection.Element == Element {
        var result = [GroupKey: ValueCollection]()

        for value in self {
            collectionReducer(&result[try keyForValue(value), default: collectionInitialiser()], value)
        }

        return result
    }
}

let result: [Int: [Int]] = [1, 2, 3, 4].grouped(by: { $0 % 2 },
                                                collectionReducer: { $0.append($1); return $0 })
```

This is essentially now a functional superset of `group(by:)` and `keyed(by:)`.

You can't provide a default reducer since there's no way to generically add a value to any type of collection (per the earlier point).  So this'd have to be a special parallel version of group(by:), in addition to the 'simple' one that just hard-codes use of `Array` and doesn't require that a reducer be specified.

I'm not a fan of this approach, technically capable as it may be, though admittedly that's from a subjective standpoint.  I think it'd be more elegant, and cleaner for library authors, to evolve Swift (or the Swift standard library) to better support generalising across all collection types.

Guides/Keyed.md
```diff
@@ -0,0 +1,78 @@
...
+let fruitByLetter = fruits.keyed(by: { $0.first! })
+// Results in:
+// [
+//     "A": "Apple",
+//     "B": "Banana",
+//     "C": "Cherry",
+// ]
+```
+
+Duplicate keys throw an runtime error by default. Alternatively, you can provide a closure which specifies which value to keep:
```
wadetregaskis, line 19: Grammar nitpick: "an runtime" should be "a runtime".

Guides/Keyed.md
```diff
@@ -0,0 +1,78 @@
+# Indexed
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Keyed.swift) | 
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/KeyedTests.swift)]
+
+Stores the elements of a sequence as the values of a Dictionary, keyed by the result of the given closure.
+
+```swift
+let fruits = try! ["Apple", "Banana", "Cherry"]
```
wadetregaskis, line 9: This `try!` is on the wrong line, I think?  The array literal doesn't throw, but `keyed(by:)` might.

Guides/Keyed.md
```diff
@@ -0,0 +1,78 @@
...
+| Java          | [`toMap`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/util/stream/Collectors.html#toMap(java.util.function.Function,java.util.function.Function)) |
+| Kotlin        | [`associatedBy`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/associate-by.html) |
+| C#            | [`ToDictionary`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.todictionary?view=net-7.0#system-linq-enumerable-todictionary) |
+| Ruby (ActiveSupport) | [`index_by`](https://rubydoc.info/gems/activesupport/7.0.5/Enumerable#index_by-instance_method) |
+| PHP (Laravel) | [`keyBy`](https://laravel.com/docs/10.x/collections#method-keyby) |
+
+#### Rejected alternative names
+
+1. Java's `toMap` is referring to `Map`/`HashMap`, their naming for Dictionaries and other associative collections. It's easy to confuse with the transformation function, `Sequence.map(_:)`.
+2. C#'s `tooXXX()` naming doesn't suite Swift well, which tends to prefer `Foo.init` over `toFoo()` methods.
```
wadetregaskis, line 69: Typo:  `tooXXX()` -> `toXXX()`.

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,86 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public struct KeysAreNotUnique<Key, Element>: Error {
+  public let key: Key
+  public let previousElement: Element
+  public let conflictedElement: Element
```
wadetregaskis, line 15: Subjective nit:  I think "conflictingElement" sounds better (more natural).

In this case, in context, it's perfectly clear either way.  The preference comes from the difference between "conflicted" and "conflicting" in a much more broad, general sense.  Technically "conflictedElement" could mean "the element that is conflicted, as in has conflicting feelings".  Nonsensical here, I know, but still it "feels" better to say "conflicting".  English is a silly language. 😆

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,86 @@
...
+
+  public init(key: Key, previousElement: Element, conflictedElement: Element) {
```
wadetregaskis, line 17: I might be missing something really obvious, but… why is an explicit initialiser necessary?  Doesn't Swift implicitly generate a memberwise initialiser?

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,86 @@
...
+    self.key = key
+    self.previousElement = previousElement
+    self.conflictedElement = conflictedElement
+  }
+}
+
+extension Sequence {
+  /// Creates a new Dictionary from the elements of `self`, keyed by the
+  /// results returned by the given `keyForValue` closure. Deriving the
```
wadetregaskis, line 26: It might help to say which specific error type is thrown.  Plus this can be moved to the formal `throws` DocC field, a la:

```swift
/// - Throws: `KeysAreNotUnique ` if two values map to the same key (via `keyForValue`).
```

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,86 @@
...
+        throw KeysAreNotUnique(key: key, previousElement: previousElement, conflictedElement: element)
+      }
+    }
+
+    return result
+  }
+
+  /// Creates a new Dictionary from the elements of `self`, keyed by the
+  /// results returned by the given `keyForValue` closure. As the dictionary is
+  /// built, the initializer calls the `combine` closure with the current and
```
wadetregaskis, line 51: `resolvingConflictsWith` (or `resolve` perhaps?) now, not `combine`.

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,86 @@
...
+    for element in self {
+      let key = try keyForValue(element)
+
+      if let oldValue = result.updateValue(element, forKey: key) {
+        let valueToKeep = try resolve(key, oldValue, element)
+
+        // This causes a second look-up for the same key. The standard library can avoid that
+        // by calling `mutatingFind` to get access to the bucket where the value will end up,
+        // and updating in place.
+        // Swift Algorithms doesn't have access to that API, so we make due.
```
wadetregaskis, line 78: Typo:  `make due` -> `make do`.

Sources/Algorithms/Keyed.swift
dnadoba: The implicit generated initialiser is `internal` but this one is `public`. Also, you probably want to mark the init `@inlinable` because it is generic and otherwise will not spezialise outside of this module.

Sources/Algorithms/Keyed.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    for element in self {
+      let key = try keyForValue(element)
+
+      if let previousElement = result.updateValue(element, forKey: key) {
+        throw KeysAreNotUnique(key: key, previousElement: previousElement, conflictedElement: element)
+      }
+    }
+
+    return result
```
LucianoPAlmeida, line 46: Would it make sense to have this in terms of the other overload? e.g.
```suggestion
    return try self.keyed(by: keyForValue, resolvingConflictsWith: { key, previousElt, conflictedElt in 
       throw KeysAreNotUnique(key: key, previousElement: previousElement, conflictedElement: element)
    })
```

Guides/Keyed.md
amomchilov: Not a nitpick at all!

Sources/Algorithms/Keyed.swift
amomchilov: > "the element that is conflicted, as in has conflicting feelings"

Maybe the element is feeling a little down. Have you tried asking it? 😅  /s 

> but still it "feels" better to say "conflicting". English is a silly language. 😆

Agree!

Sources/Algorithms/Keyed.swift
amomchilov: I wasn't sure about `@inlinable`.

> otherwise will not spezialise outside of this module.

Since this is kind of an edge case, is the perf improvement worth it relative to the code size cost of specializing it? What's the SOP for deciding on that?

Sources/Algorithms/Keyed.swift
amomchilov: Gah I thought I got all of these. Nice catch

Sources/Algorithms/Keyed.swift
amomchilov: TIL

Sources/Algorithms/Keyed.swift
amomchilov: Oooo fantastic! This was a case of path-dependence, it wasn't possible before, but now that it's two functions, it's available and much nicer!

Guides/Grouped.md
amomchilov: Very interesting findings, thanks for sharing this with me! How do you think it would be best to communicate these in the library? Perhaps an addendum to the new md files I'm adding in this PR?

> but you still need a protocol that all collections support which specifies some kind of append method, and that doesn't currently exist.

Oh wow, I never noticed this, but you're right!

> It is possible to support arbitrary collection types, but only by having the caller provide a reducer explicitly in order to work around the aforementioned limitations, e.g.:

This is what I had in mind (or use a new protocol with method requirements, instead of two separate closure params).

This is more general, because the groups might not be collections at all. The histogram example comes to mind, where you'd want something like:

```
[1, 2, 3, 4].grouped(
    by: { $0 % 2 },
    collectionInitialiser: { 0 }, // not a collection, but w/e
    collectionReducer: { counter, _ in counter += 1 },
)
```

If there were a protocol for this, then there might be some `CountingCollector` that could be used.

amomchilov:
Thanks for the feedback everyone! I'll try to incorporate it all this weekend

Guides/Grouped.md
wadetregaskis: I'm not sure how best to summarise that - obviously what I wrote, as a kind of deep-dive on this tangent, is way too much for these documents.  Maybe it suffices to say something like "This would require either (a) a new protocol, supported by all standard library collection types, which represents an initialisable object that can be added to, or (b) a variant method which accepts a custom 'reducer' closure".  If you want to say anything about it at all.  Perhaps this thread here on the pull request is sufficient documentation for posterity (which could be hyperlinked to from the markdown file).

Having used Java's version of this a few times in real-world code, my main comment is that these sorts of APIs - groupby etc - should never _require_ explicit specification of the return types (or a custom reducer).  They can _allow_ it, through overloads or optional parameters, but it needs to work effortlessly for the vastly most common case of just returning a Dictionary of Arrays.  A lot of Java APIs explicitly require you to pass a collector which is almost always `Collectors.toList()`… it gets very repetitive, and verbose.

wadetregaskis: Also, as a counter-argument (Devil's advocate style) for the use of this custom-reducer `grouped(by:)` variant, you can achieve the same thing with:

```swift
[1, 2, 3, 4].reduce(into: [Int: Int]()) { $0[$1, default: 0] += 1 }
```

(and the explicit typing of the `Result` type can be omitted in some cases where it's implied by surrounding context, making this even terser)

Personally I don't mind if a language or library provides multiple ways to achieve the same goals, but I'm not sure a fancier `group(by:)` adds any benefit over the existing `reduce(into:_:)` approach.

That said, you can implement `group(by:)` using `reduce(into:_:)` too, e.g.:

```swift
[1, 2, 3, 4].reduce(into: [Int: [Int]]()) {
    $0[$1 % 2, default: []].append($1)
}
```

…yet I still feel like a purpose-built `group(by:)` is worth having.  Perhaps it's technically just a "convenience", a bit more ergonomic - but it feels like it's worth it for this simple, common case.  I guess I'm just not sure where to draw the line between a more powerful `group(by:)` and just having people fall back to `reduce(into:_:)` for uncommon needs.

Counter-counter-argument:  even though `reduce(_:_:)` and `reduce(into:_:)` _can_ be used for a huge variety of tasks, I usually find them hard to interpret when reading the code.  A large number of `Sequence` methods can be implemented using just the reducer methods - e.g. `allSatisfy`, `min` / `max`, and even `map`! - yet we still have all those purpose-built methods.  And I think Swift is better for them - they make code much more readable, even if they're technically redundant.

Sources/Algorithms/Keyed.swift
LucianoPAlmeida: > Since this is kind of an edge case, is the perf improvement worth it relative to the code size cost of specializing it? 

IIUC `@inlinable` will just make the body of function available for specialization outsize the module, but at the end is the optimizer which will decide based on a cost model if is going to inline or not. e.g. if user compiles with `-Osize` it may not inline that call at all even though is marked as inlinable. For given a hint that we always want to inline, we can use `@inline(__always)` together with `@inlinable`. 

> What's the SOP for deciding on that?

With that said, I think is always worth to add `@inlinable`. Specially in this case as code size impact would be not significant. But I also see why it wouldn't worth it

Guides/Grouped.md
amomchilov: > I guess I'm just not sure where to draw the line between a more powerful `group(by:)` and just having people fall back to `reduce(into:_:)` for uncommon needs.

Likewise. I have conviction about the base `group(by:)` and its semantics, but I don't know how far its extensibility should go.

>Counter-counter-argument: even though `reduce(_:_:)` and `reduce(into:_:)` can be used for a huge variety of tasks, I usually find them hard to interpret when reading the code. A large number of `Sequence` methods can be implemented using just the reducer methods - e.g. `allSatisfy`, `min` / `max`, and even `map`! - yet we still have all those purpose-built methods. And I think Swift is better for them - they make code much more readable, even if they're technically redundant.

Heh I have an entire little blog post about this: https://github.com/amomchilov/Blog/blob/master/Don't%20abuse%20reduce.md

Sources/Algorithms/Keyed.swift
amomchilov: Interesting, there are no other thrown errors in the standard library, or even `swift-algorithms`. Should this be a documented error that we expect users to catch, or should it be treated more like a softer version of a `fatalError`?

Sources/Algorithms/Keyed.swift
amomchilov: Is this something you'd like to commit? You deserve to have your name on your findings :)

Sources/Algorithms/Keyed.swift
wadetregaskis: I think it should be documented, mainly because callers need to be able to reliably distinguish between errors thrown by this library method itself vs deeper errors from the closures, because they might want to handle those cases differently (e.g. distinguish to the user that the problem is duplicate keys vs an invalid value, or failure to fetch some additional data from the network, etc).

As to this being novel for this library, that might just be coincidence?  Perhaps swift-algorithms just doesn't happen to contain a lot of cases where there is an error condition (that can't otherwise be handled implicitly, such as returning nil).

Looking at [all the uses of `fatalError` in this library](https://github.com/search?q=repo%3Aapple%2Fswift-algorithms%20fatalError&type=code), they're all related to indexing errors.  Those are a different breed, I think - they're more likely to represent actual code bugs, for a start.

Pragmatically, perhaps the key decision point is how much work it is to pre-emptively validate the data vs just catching an exception.  For subscripting, for example, it's arguably easy & fast enough to just check if your index is within range first.  A similar argument can be made for why the basic arithmetic operators in Swift will just crash on overflow / underflow / etc rather than throw exceptions.  But for more complex operations - like `group(by:)` and `keyed(by:)`, pre-emptively checking for duplicate keys means basically reimplementing those methods.  I don't think that's reasonable to expect of callers.

#### Sidebar on crashing philosophy

There's a quiet philosophical debate within the Swift community about how to handle errors.  Some folks think it's preferable to just crash (and thus save API users from having to deal with exceptions), while others (myself included) prefer to give callers control over whether they crash or handle errors in some other manner.

Since crashing can be achieved with just a suitably-placed `try!`, there's very little added cost for people that want their application to crash instead of having to handle the exception.

For most use-cases, crashing just isn't acceptable (e.g. many server applications, most end-user applications, etc).  Bad input data should just fail the relevant, specific operation, not the entire application.  e.g. even if it's the Save operation that's broken by the exception, letting the application keep running at least gives the user a chance to save their data in other ways (e.g. copy-pasting it into another application, or dragging it out to a file in the Finder, etc).

More broadly, crashing is almost always an unsafe operation - who knows what things in the system, like databases or other data stores on disk, are in an invalid state right at that moment.  The stack needs to be gracefully unwound in order to ensure persistent state is not corrupted.

Guides/Grouped.md
amomchilov: Is this research something you'd like to commit? You deserve to have your name on your findings :)

Guides/Combinations.md
```diff
@@ -86,7 +86,7 @@ call to `CombinationsSequence.Iterator.next()` is an O(_n_) operation.
 ### Naming
 
 The parameter label in `combination(ofCount:)` is the best match for the
-Swift API guidelines. A few other options were considered:
+[Swift's API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/). A few other options were considered:
```
stephentyrone, line 89: I'm fine with this change and the previous change, but they should be in a separate PR.

amomchilov: Getting one PR merged has been difficult enough, I'm apprehensive to open a second just for something so tiny and low-value 😅

stephentyrone: I am happy to rubber-stamp that one immediately =)

amomchilov: Done! https://github.com/apple/swift-algorithms/pull/210

Could you please review this PR next? :D

stephentyrone: Yeah, @natecook1000 and I will review this week

natecook1000 (commented):
Thanks again for your work on this, @amomchilov, and especially for your perseverance in seeing this addition land!

Let's use a slightly revised solution for the `keyed(by:)` overload: this single-argument version should act like the code that it would typically replace – that is, keeping just the last element if there are any duplicated keys produced by the `keyForValue` closure. Especially with the potential upcoming advent of typed throws, I'd prefer to keep the existing system of only rethrowing errors from user-provided closures, instead of throwing new errors directly from the algorithms library.

```swift
let fruits = ["Apricot", "Banana", "Apple", "Cherry", "Avocado", "Coconut"]
let keyed1 = fruits.keyed(by: { $0.first! })
// keyed1 == ["A": "Avocado", "B": "Banana", "C": "Coconut"]

// equivalent to:
var keyed2: [Character: String] = [:]
for fruit in fruits {
    keyed2[fruit.first!] = fruit
}
```

This behavior matches that of the other languages that have this feature already (except for Java, which predictably throws an exception), and seems like it will lead to less surprise than having people reckon with the potential thrown error.

It _doesn't_ exactly match the behavior of of the `Dictionary(uniqueKeysAndValues:)` initializer, but I think that's fine, given the presence of the key-producing transformation. For those who would like to customize the value selection (or throw an error or trap), there's always the second `keyed(by:resolvingConflictsWith:)` method.

amomchilov:
@natecook1000 That sounds good by me! I'll try to rework this on the weekend :)

amomchilov:
@natecook1000 @stephentyrone Hey guys

I rebased, and my latest commit changes from the original "throw on collision" behaviour to the new "keep latest" behaviour.

natecook1000:
@swift-ci Please test

natecook1000 (approved):
🎉

stephentyrone:
Thanks for your persistence on this one, @amomchilov!

PR #203, opened 2023-09-09
Add a non-mutating lazy replaceSubrange

karwa:
Sometimes, you want to insert/remove elements from a collection without creating a copy and mutating it.

It's kind of a generalisation of `chain`. And maybe `joined`. I still need to do the guide document, but I'm creating the PR now for discussion purposes. In particular:

- Is this a good candidate for inclusion in the project?

- What should it be named? 
  `x.lazy.replacingSubrange(i..<j, with: y)` isn't _amazing_, but draws strong parallels with `replaceSubrange`.
    I could also see "overlaying", as in `x.overlaying(i..<j, with: y)`. Any other ideas?

- Does anybody know of similar features in other languages?
   For the guide doc. I tried searching but couldn't find anything.

### Checklist
- [X] I've added at least one test that validates that my change is working, if appropriate
- [X] I've followed the code style of the rest of the project
- [X] I've read the [Contribution Guidelines](../blob/main/CONTRIBUTING.md)
- [] I've updated the documentation if necessary


karwa:
So, when [checking this out using Godbolt](https://swift.godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AKA7nioFSS%2BsgJ4BlRugDCqWgFcWDPU4AyeBkwAOU8AI0xiPQAHVAVCewZXDy9o2PiBf0CQlnDIritMGzsBIQImYgIkz29860xbBNLygkzgsIi9BTKKqpTa7paAtpyOrgBKK1R3YmR2DgBSDQBBKncGZABqAkwuiAB9DfUNgCVMKINkAOAhd1DiQ2BMJKKE%2BYBmZ3mAVgAhAElBN8ACKkDbff6Ar5A97YMZggDsP0WSw2qI2ADdyhs8GC3kDDgA6LrNAFYVTItEbEwIOiYbG45yEpykzCqBFI5aU1GhC4AawQbkwEHJvzwwLGFK5qgJ/GILBZqggTAsHTBACYAGx4CWctHzeHQ5b6w1LZEAejNG1W6y2OwI%2B0OIA2S2I9wAnu9nACCDC4fqOUsLWjZRjsQxCQYAF5ugnEM4XK43O4PIVfAkEz1cDSgswEJAbSQadOer4aP2I82Wyk8pjIfmCiDonWBqsIk1B43I5EAAT2cVojwYe2YbAg8zVaprdYF9HHaub1s2U/r9H2Tu9za7RuWFve0Lxe4AtMeT6ez%2BeL5er9eb6e9/e8bud2bKxsACo0hQbBRTGZ0mj0NiX5RM0GyoFQWwIHSQhmBYzoDiQhAICwX6oFEjDfr%2BswbFExCoNo9Q%2Bs%2Br6uFEbrEHgwAIAQGwQMgcJqhojHOlE5x0gCyAEhshjoJB0GwTRuH4YR3HuHmJAKK%2B/izAwSi8WsWDECxtZQRs0mMEoGJqgSGhUkhJxrHYbBqXgybEG6GzYKosxRMUDCVq%2BQiYHS1EEFECggBapjmAQBIkMAZq%2BH8zjYEEQjYASBCqDRIa0HgMmaQEsosEwdkOcRB6ZUCt45bleX5Q%2B%2B5Ak%2BprLKy2yyQkalMNGzyEVV/pbiiGzdgEcXMDymCSlEtxxZsi4bHG5y1omtz3EYXUfKcw2zGwgiSW82BjrqqIHAoY0pk6xwpp6Cowjm%2BmBCY2D0HNBAeSc8a1pgZ2SbqcKHjCl0zaNyYTXVDQCJ6PxMEooLTRcN2MOdT2NStz0JkYSbjY8H12RAoS/ewGyFED82gkNgO3U6R0nWj50Y1dszoNtE1Outb2PM2qKdkaBpNT1PLxd%2BBDEO4tgQyNUMbe9bgvF9Hw/X9nOzcDC3YMi1IRHSQvI3DCT/UT%2BMXfLAigrLBJ42duJFSLyua6dwPsk1qLdu4ChMJ1ABieHygw7VdbqATbMQzC0BiWKI0oTqyybLXm5b9A26gdsO5KzsRG7HtKZj13Y3rt1%2B2bFvW7bAJh07giR2I0eDUrJObScO2C0jBJ7YtSdtQEgeO81EeuznASEAjSM%2B0jhMzcrW1K7dHeAwXZNFxN32l%2BXsLG%2BD1hUASXt0nuGyz5KqJT7GPdizreed4nk%2BFNPsfE6Tjwb/vmAD48kq06a9Pbks5UaVVANc9cPOw3z9UOBsqvhmDpXNYzfVbDdOhMQeBfqWUNoIDeGstbAz9v/ZmXQ2YcwVE6VwLAQL3E6hPZqNNwbJxrsHUOQwl5hhdlHRgngNgAHV7isVPtgrkGxRCaVnhADWCpqZcmYXSE%2BZ0ICP1FvNMuDAyScJptfX%2BlILSWg/HSBSERaBuiuAvJGZpeHrwCGSDM4NpEkPwanEO6diHg3rlHTESkTC0PQugJ0NCmB0PQH7VE0j3yqVnmGfAswvzUniggDYCAmDojpOEDCJ90DaJwRsXReCA4GKIYEEhpic7mM3v3Q%2ByN0kjyUMIskMInEtSrh1WcJis4N3dk3e0lj7HWNsVY0%2BfdrpnwycXZw7CRGsl9AwrkK8qkOI3r06xJDl671Xp3Jpx986ZPBpfPRhSa4kPgZsYkdh%2BprE2J6WitAEAXScrQKgGNtlOl2VQB6T0fioDcF0yk3kCDID8RALZCgCQDPqYNbZzy6noHLAGRh3DaIzyRg8zANE1gvPQL4TAFgJgbABUoIFIKGBguOJRaiYwxggCGWiOMBBpjhlBZ8iFcENn4uqafZFVEiKRNRH8iAoysbA3hRsElDjCUEGhXSuODL6AIqRSitl6LMWomxbipliKCWQposSsVpKSZ8sxTS2FQpuWirBay9l6jBD7DRRi8GlJhWuxVeKolHwNiPI5QfFMBJaCoBMBEc5Cl5VI3%2BRq%2B0ex2WsOVcy6x5LUUCt1Vi4FIqwCQC9WSvlDJTXvLCekq1Nq7VTBEWIvUEjKQzLwXMzqCzeoILKCsq0aydbzweYcjYxyDk7N3qcxaGxzmXJ/owoMQRUAbECPQogTCoJ1g2GADg0aUw9tBOEUQ5s6SEA2OgVAOwW2oBovgKgVAIgL2BbajCmj4pTqoLbPi35WB0jQLQfmDAImMJuXczZ7ywXlo%2BTK75jrNK0o9cCw1Mq1WgkVYy0NsqKXasFXnHFBrP2ssLXiZ9DifWUt%2BU62lLqP3SpZRK9Va9NWerg96vlP7/VCsDQB1Dp8gOFVA2hilmKsBUCYO4WgBAdVUoDf%2B8MZHaBKExZfVNKa2xdgkciO%2BlUBAi1ejDJ4b9PrfwrDfU2Gbil11KVHAaKVeSYAVA6GIcQ7Jt2yRwjYj1q0KiuRsRTYKnTvuU%2BkBg0K%2B2D3M1TC%2BnHdStXttXTNmcyE51k0weTimDjGdUwnMWOTWRVuwHp9pbJ61on058wzMGvMJDM5MwulnMCcLTc1Oz7V5m6kWbnYkFQUFBbJLp8wKjsnZYIDpgjCXY22uIPakRunUSFZdU8vAChsDoIIBZULjD9Xhjkwp4LEAKvuDodVhNXzmNsb1dhnrbm%2BtkgG0h86RIBgcJISx2jIreuKdnktkkwWks2ZSxJ2uqJMspOZMF9cwWCsQQq9aqrNXeLBvntt87%2BX9SfHhIyRrBJmutdsh10TNHusbE2/1l7IiVvTIm3%2BjbM3FPfde/56zJpJSpYc5Jk72bNgpLQIZS7krAfVlLvgYk6whSbpDk6bbJWFSgiIE6W7caRsKSTeOH4qTOVCNx4CaZap2fbZJ2UMn1BbYM7ixNAkQ30LM5EXT1AVPS6I9UPtlHtmjvdSx/mm0mjWRKhVEpCgeX/OaaemVwnaJT1%2BLwNevpnWmFOvfSk2eHDqOMO28lRTyoXZOnHBqZ3e3MWFf929rKHOLUS7u/GhSdW0QNYW01lrbWAc/MYetg1oO5uDeGw9pNrGTSp5h%2BnuH/Xvs04D/6tbk26Mg%2BL3N4PSOxNoj%2Bea/GjYsQupd5i77Hv%2Bte9VL7jv5eaNx7GTGyPMvHtgGe4riHwWGmCNK1d8r8eCRK5j1h6vGfddZ%2Blzn8b%2BeuTA634qQfojVtseS%2BJ%2BzRTjs4U1xTBQyAKK2QgNFj%2BGmtOBZgZAu3lucJpDsht2sV02b0fRonr2Vx939WB0Xl%2BAgOBBIWbxg2VVP38ygKBymzD2VnBFQJFAP3EQPxmQNA4AmFoE4C%2BF4G8A4C0FIHlw4Bgh8kwmmGwnHDeB4FICo2oJIImF5BADeAAE4CQuANRJBJB%2BCNAAAODUYQrgCQiQ/QTgSQSgzQXgOg3gDybMTgrQCYOAWAJANAdBWkMgCgOiEOKIIwkAbyCwBgPlRRUgLAdEddAANTwEwBMAAHl0IqD2CAIXYPIEYVDSBQhq5zJOB2DgjmBzJ3DQgCJbAwjeADCzp3D7Y3RAisBQh3BgBnAxBGN4j7CboHhxAuD7C8A4wGggkPJijWR6gxI5gaCI4yDii4ozI3RXAsBAjWY8AWB4iJgqADBgAFAXC3DPDGA8j%2BBBARAxB2ApAZBBBFAVB1BijdB8gDAjBLCBJ9BTIPJIAJg0I7JKjDx3DVCglXQ8AsBtiIAJg6hhNHARFegahSA/AhhshchUgVMEh7i3iTNWgXjRgChD0mgeg3BqhOhUZhNATBgsh2g8grABhPj%2BhmgfjoSJArisJpjSDyDlDii6DDgpDDwRDvwBJNgbCKVFFaJcBCASB1Q2CxheAtDuDSBeCNQiwJC3gvhSx%2BCuA3hWSJCuBGIvgFCOAlDSAqCaC1CrAQBNCVC0VBS1QsSxTOA6TpSJgTi4gHBJAgA%3D%3D), I find that the compiler produces shockingly good code. To the extent that, while I initially thought it might be a bit niche, I now wonder if it could be an interesting alternative to mutating operations for a lot of other applications - particularly SwiftUI applications.

Imagine I have some data, perhaps retrieved from the network or stored in some global datasource object. Let's say they're some kind of product, perhaps shoes, and I'm displaying them in a list/grid. I also have a boolean, telling me whether I should display a special call-to-action item in the list. This is a fairly standard thing to want.

The actual data that I want to use to drive my list/grid is composed from those two elements: the list of products and the boolean.

```swift
class ProductsListViewModel {
  var rawProductsList: [Product] = ...
  var showCallToAction: Bool = ...

  var viewItems: ??? // built by combining rawProductsList with showCallToAction
}
```

SwiftUI tends to emphasise computed properties for these kinds of things -- while I _could_ make `viewItems` a stored Array-type property, I'd then need to make sure it's always up-to-date. Also, if the products list is a reference to non-owned data (such as a `Binding`), it may update even when this particular view isn't on-screen; in that case, we might be recomputing the `viewItems` unnecessarily.

Computed properties, on the other hand, are always up-to-date, taking account of the latest list of products and the latest state of the boolean. With `@Observable`, these kinds of data dependencies will even be tracked automatically.

However, within the implementation of the computed property, we must make a copy of the products list in order to inject elements in to it. The lack of lazy insertions also means we have to perform eager transformations such as `map`.

```swift
class ProductsListViewModel {
  var rawProductsList: [Product] = ...
  var showCallToAction: Bool = ...

  var viewItems: some Collection<ListItem> {
    var result = rawProductsList.map { .product($0) }
    result.insert(.callToAction, at: min(result.count, 4))
    return result
  }
}
```

This is unfortunate, since `viewItems` will likely be called in a SwiftUI `body` property, meaning the `map` copy, and potential second copy and element shuffling caused by the `insert`, will happen in this performance-critical function.

With lazy insertions, we can avoid that:

```swift
class ProductsListViewModel {
  var rawProductsList: [Product] = ...
  var showCallToAction: Bool = ...

  var viewItems: some Collection<ListItem> {
    let result = rawProductsList.lazy.map { .product($0) }
    let ctaPosition = result.index(result.startIndex, offsetBy: 4, limitedBy: result.endIndex) ?? result.endIndex
    return result.replacingSubrange(ctaPosition..<ctaPosition, with: CollectionOfOne(.callToAction))
  }
}
```

`viewItems` has the same contents as before, but now we're able to use a lazy `map` transformation on the products list, and our insertion of the call-to-action prompt also won't copy or shuffle elements around. I'd expect it to be cheaper to call this from within a SwiftUI `body` function.

Iterating such a collection is more expensive than iterating an array, but because the compiler does such a good job with the generated code, there's a very good chance that it's still a net positive.

Unfortunately, it is also more awkward to use. Which is why another idea that I have WRT naming is to create a `.overlay` namespace struct (similar to `.lazy`), on which we can implement similar convenience methods to `RangeReplaceableCollection`. So people could write things like:

```swift
result.overlay.insert(.callToAction, at: ctaPosition)
```

(Array obviously has integer indexing, which is supremely convenient. We can't quite match that, but we could perhaps add some convenience APIs such as `insert(_: Element, atOffset: Int)`)

karwa:
One issue is that, if you're only conditionally applying a replacement while returning an opaque type, you need to perform a no-op replacement which ensures both branches have the same type. e.g:

```swift
if showCallToAction {
  return result.overlay.inserting(.callToAction, at: min(4, result.count))
} else {
  // Must return an OverlayCollection<Base, CollectionOfOne<ListItem>> which only contains the base elements.
  // That's not trivial.
}
```

What we can do instead is to (internally) make the overlay elements optional (i.e. the stored property `overlay` would become an `Optional<Overlay>`). That means we could always construct a no-op overlay wrapper with any overlay type.

To expose that feature, we could add a conditional overlay function:

```swift
extension Collection {

  @inlinable
  public func overlay<Overlay>(
    if condition: Bool,
    _ makeOverlay: (OverlayCollectionNamespace<Self>) -> OverlayCollection<Self, Overlay>
  ) -> OverlayCollection<Self, Overlay> {
    if condition {
      return makeOverlay(self.overlay)
    } else {
      return OverlayCollection(base: self, overlay: nil, replacedRange: startIndex..<startIndex)
    }
  }
}
```

Allowing usage like:

```swift
var viewItems: some Collection<ListItem> {
  let result = rawProductsList.lazy.map { .product($0) }
  return result.overlay(if: showCallToAction) {
    $0.inserting(.callToAction, at: min(4, $0.count))
  }
}
```

lgreydev:
replaceSubrange  -  Nice!

CTMacUser:
For the single-element removal method, isn't "removingSubrange(position..<position)" wrong? The range above specifies zero elements. You'll need "position ..< NEXT(position)" instead, where NEXT is replaced with the appropriate code.

karwa:
@CTMacUser Yes, good catch!

CTMacUser:
`OverlayCollection.makeIndex(_:)` can't work if the source collections use the same `Index` type, because then the two overloads would have identical signatures. Possible fix: explicitly use the words "base" or "overlay" within the methods' names.

CTMacUser:
You could add an optimized `.isEmpty`

```swift
  @inlinable
  public var isEmpty: Bool {
    base[..<replacedRange.lowerBound].isEmpty && base[replacedRange.upperBound...].isEmpty && (overlay?.isEmpty ?? true)
  }
```

karwa:
> `OverlayCollection.makeIndex(_:)` can't work if the source collections use the same `Index` type, because then the two overloads would have identical signatures. Possible fix: explicitly use the words "base" or "overlay" within the methods' names.

Yes it can, because whether we're constructing a base or overlay index is determined based on unbound generics. Specialisation happens later, and does not affect overload resolution. This is one of the ways in which Swift generics are not like C++ templates.

For example, the very first test appends an `Array<Int>` to a `Range<Int>`. Both types use `Int` as their index.

karwa:
@natecook1000 Thoughts?

CTMacUser:
In my run through of [`598946f`](https://github.com/apple/swift-algorithms/pull/203/commits/598946f0f7b47c0fadfead8b130a17e1cd490f01), that last default of `true` in `isEmpty` was never triggered. That's because the trigger would be excluding an overlay on an already-empty collection. My quick test case, appended to `testConditionalReplacement`:

```swift
    func getNoNumbers(shouldInsert: Bool) -> OverlayCollection<Range<Int>, CollectionOfOne<Int>> {
      (0..<0).overlay(if: shouldInsert) { $0.inserting(42, at: 0) }
    }

    do {
      let result = getNoNumbers(shouldInsert: true)
      XCTAssertEqualCollections(result, [42])
      IndexValidator().validate(result, expectedCount: 1)
    }

    do {
      let result = getNoNumbers(shouldInsert: false)
      XCTAssertEqualCollections(result, [])
      IndexValidator().validate(result, expectedCount: 0)
    }
```

PR #205, opened 2023-10-05
Increase the package version to 5.4

natecook1000:
The latest Swift versions appear to have a problem with the LinuxMain.swift again, which we haven't needed for quite some time. Linux test autodiscovery has been enabled since 5.4, so we can update to that version and drop the file.


natecook1000:
@swift-ci Please test

Issue #219, opened 2024-01-06
`CombinationsSequence` should be a `Collection`


dabrahams:
It's multipass (like almost every other `Sequence` in practice).  Not being a collection limits its applicability in silly ways.

natecook1000:
@dabrahams Could you give some examples of the limitations that you're running into here?

IIRC, the reason that `CombinationSequence` is only a sequence is that its iterator stores and mutates a single copy of the base collection's indices as it iterates. By contrast, a collection would need to provide an index type that either stores a separate copy of the indices (potentially very heavyweight) or stores just a position, and then recalculates the indices when used (which should be doable, but might be costly). While the `Collection` protocol doesn't explicitly rule out such a heavy index type, it's very different from the way most collections work, and having O(N) work happening in each subscript would probably be unexpected for users.

That said, `CombinationSequence` might already be falling down on the efficiency job. It looks like the iterator currently copies each combination into a new array, instead of using a collection wrapper to combine the base collection and the array of indices. Such a wrapper would be independently useful, as well (and indeed I thought we already included it):

```swift
let numbers = [10, 20, 30, 40, 50]
for x in numbers.indexed(by: [4, 2, 0]) {
    print(x)
}
// 50
// 30
// 10
```

dabrahams:
> Could you give some examples of the limitations that you're running into here?

I didn't run into a limitation, but I know such limitations exist.  It's the whole reason I wrote [Zip2Collection](https://github.com/loftware/Zip2Collection): there are algorithms that are defined only on `Collection` because it is multipass. 

> IIRC, the reason that CombinationSequence is only a sequence is that its iterator stores and mutates a single copy of the base collection's indices as it iterates.

That's no reason it can't be a collection, as you yourself [know well](https://forums.swift.org/t/pitch-make-collection-super-convenient-and-retire-sequence/12103) and  [this implementation](https://github.com/hylo-lang/hylo/blob/f5c3cfd1fda16aa530d872b60b148dfd5e2c6260/Sources/Utils/Combinations.swift) demonstrates.

> having O(N) work happening in each subscript would probably be unexpected for users.

I don't know what your N is in this case; it's certainly not the length of the resulting collection, which is what people usually mean when they say O(N).

The work of dereferencing this thing is proportional to the data footprint of the elements you get, which is no worse than `someStrings.lazy.map($0 + ".")`.

[...]

PR #220, opened 2024-01-06
CombinationsSequence: underestimatedCount == count

dabrahams:
Not doing this makes it needlessly costly to create an Array of the results.
It's a reasonable incremental improvement until you get around to #219, which would handle this automatically.

### Checklist

### DIFF
--- a/Sources/Algorithms/Combinations.swift
+++ b/Sources/Algorithms/Combinations.swift
@@ -78,6 +78,10 @@ public struct CombinationsSequence<Base: Collection> {
       binomial(n: n, k: $0)
     }.reduce(0, +)
   }
+
+  /// The total number of combinations.
+  @inlinable
+  public var underestimatedCount: Int { count }
 }
 
 extension CombinationsSequence: Sequence {
--- a/Tests/SwiftAlgorithmsTests/CombinationsTests.swift
+++ b/Tests/SwiftAlgorithmsTests/CombinationsTests.swift
@@ -15,57 +15,64 @@ import Algorithms
 final class CombinationsTests: XCTestCase {
   func testCount() {
     let c = "ABCD"
-    
-    let c0 = c.combinations(ofCount: 0).count
-    XCTAssertEqual(c0, 1)
-    
-    let c1 = c.combinations(ofCount: 1).count
-    XCTAssertEqual(c1, 4)
-    
-    let c2 = c.combinations(ofCount: 2).count
-    XCTAssertEqual(c2, 6)
-    
-    let c3 = c.combinations(ofCount: 3).count
-    XCTAssertEqual(c3, 4)
-    
-    let c4 = c.combinations(ofCount: 4).count
-    XCTAssertEqual(c4, 1)
-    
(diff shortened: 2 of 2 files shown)


Sources/Algorithms/Combinations.swift
```diff
@@ -78,6 +78,10 @@ public struct CombinationsSequence<Base: Collection> {
       binomial(n: n, k: $0)
     }.reduce(0, +)
   }
+
+  /// The total number of combinations.
+  @inlinable
+  public var underestimatedCount: Int { count }
```
pyrtsa, line 84: Doesn't this conflict with the complexity requirement of `Sequence.underestimatedCount`?

```swift
  /// - Complexity: O(1), except if the sequence also conforms to `Collection`.
  ///   In this case, see the documentation of `Collection.underestimatedCount`.
  var underestimatedCount: Int { get }
```

The requirement is relaxed for non-random access collections, so not a problem for #219.

dabrahams: Hum, indeed.  What is the complexity of your `count`?  If it's log(N) IMO this is acceptable; complexity requirements are somewhat fungible (e.g. `first` on a lazy filtered collection is arguably not O(1)).  Otherwise, I suppose not.

Oh, but they totally broke the complexity bound of `underestimatedCount`.  It was O(n) before ea49459b7594.  What a debacle.

dabrahams: Actually, that O(1) constraint is meaningless as outlined [here](https://github.com/apple/swift/issues/70755).  You should ignore it.

natecook1000:
@swift-ci Please test

Tests/SwiftAlgorithmsTests/CombinationsTests.swift
```diff
@@ -15,57 +15,64 @@ import Algorithms
...
-    
-    let c14 = c.combinations(ofCount: ...3).count
-    XCTAssertEqual(c14, 15)
+    check(cHas: 0, combinationsOfLengths: 5...10)
     
-    let c15 = c.combinations(ofCount: 0...).count
-    XCTAssertEqual(c15, 16)
+    check(cHas: 16, combinationsOfLengths: 0...)
+    check(cHas: 15, combinationsOfLengths: ...3)
+    check(cHas: 16, combinationsOfLengths: 0...)
```
natecook1000, line 75: Duplicated test?

dabrahams: 🤷‍♂️ It was there before my change.
```suggestion
```

natecook1000: Huh! So it was.

natecook1000:
@swift-ci Please test

Tests/SwiftAlgorithmsTests/CombinationsTests.swift
dabrahams: Things beome apparent when you remove boilerplate.

PR #224, opened 2024-02-18
Add `ends(with:)`

amomchilov:
This PR adds an end-counterpart to the built-in [`Sequence.starts(with:)`](https://developer.apple.com/documentation/swift/prefixsequence/starts(with:)).

### Checklist


Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,95 @@
...
+  ///   collection and the length of `possibleSuffix`.
+  // Adapted from https://github.com/apple/swift/blob/d6f9401/stdlib/public/core/SequenceAlgorithms.swift#L235-L252
+  @inlinable
+  public func ends<PossibleSuffix: BidirectionalCollection>(
+    with possibleSuffix: PossibleSuffix,
+    by areEquivalent: (Element, PossibleSuffix.Element) throws -> Bool
+  ) rethrows -> Bool {
+    var possibleSuffixIterator = possibleSuffix.reversed().makeIterator()
+    for e0 in self.reversed() {
+      if let e1 = possibleSuffixIterator.next() {
```
amomchilov, line 83: Is there an efficient way to implement this without needing backwards iteration of the `possibleSuffix`?

The only way I can think of is basically a [substring search](https://en.wikipedia.org/wiki/String-searching_algorithm), and checking if it ends up match against the very end.

If such an algorithm existed, it would allow us to loosen the requirement of `possibleSuffix` from `BidirectionalCollection` down to just `Sequence`.

Guides/EndsWith.md
amomchilov, line 1: One of the pages of the "Guides" should link to this, but I'm not sure which.

I don't see any predicate methods like this elsewhere in the package. Perhaps we should start a new `Guides/Predicates.md` category?

Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,95 @@
...
+        if try !areEquivalent(e0, e1) {
+          return false
+        }
+      }
+      else {
+        return true
+      }
+    }
+    return possibleSuffixIterator.next() == nil
```
amomchilov, line 92: Should we just replace this full re-implementation and delegate to `starts(with:)`?

```suggestion
    self.reversed().starts(with: possibleSuffix.reversed())
```

Sources/Algorithms/EndsWith.swift
amomchilov: If we loosen the requirement to just `Collection` (so that we have a known `count`), then we can probably iterate from the middle (starting at `self.index(self.endIndex, offsetBy: possibleSuffix.count`), and matching until the end. I'm not sure if this is better though, because we need to do this (potentially-non-constant-time) index math.

natecook1000: Right … I don't imagine that there are many cases where you have a bidirectional collection but you can't make the suffix you're searching for bidirectional.

Sources/Algorithms/EndsWith.swift
natecook1000: Seems like this is what that delegation would do! We could start with the delegation and see if we need to revise it based on performance.

Guides/EndsWith.md
natecook1000: > any predicate methods like this
Can you say more about what you mean here?

Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,84 @@
...
+  /// collection are the same as the elements in another collection.
+  ///
+  /// This example tests whether one countable range ends with the elements
+  /// of another countable range.
+  ///
+  ///     let a = 8...10
+  ///     let b = 1...10
+  ///
+  ///     print(b.ends(with: a))
+  ///     // Prints "true"
```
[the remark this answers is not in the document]
amomchilov: @natecook1000 I copied this doc from `starts(with:)` and modified it. I think if we change this, we should `starts(with:)` to match.

Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,84 @@
...
+  /// - Parameter possibleSuffix: A collection to compare to this collection.
+  /// - Returns: `true` if the initial elements of the collection are the same as
+  ///   the elements of `possibleSuffix`; otherwise, `false`. If
+  ///   `possibleSuffix` has no elements, the return value is `true`.
+  ///
+  /// - Complexity: O(*m*), where *m* is the lesser of the length of the
+  ///   collection and the length of `possibleSuffix`.
+  // Adapted from https://github.com/apple/swift/blob/d6f9401/stdlib/public/core/SequenceAlgorithms.swift#L281-L286
+  @inlinable
+  public func ends<PossibleSuffix: BidirectionalCollection>(
```
amomchilov, line 45: Adapted from https://github.com/apple/swift/blob/d6f9401/stdlib/public/core/SequenceAlgorithms.swift#L281-L286

Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,84 @@
...
+  ///     are equivalent; otherwise, `false`.
+  /// - Returns: `true` if the initial elements of the collection are equivalent
+  ///   to the elements of `possibleSuffix`; otherwise, `false`. If
+  ///   `possibleSuffix` has no elements, the return value is `true`.
+  ///
+  /// - Complexity: O(*m*), where *m* is the lesser of the length of the
+  ///   collection and the length of `possibleSuffix`.
+  // Adapted from https://github.com/apple/swift/blob/d6f9401/stdlib/public/core/SequenceAlgorithms.swift#L235-L252
+  @inlinable
+  public func ends<PossibleSuffix: BidirectionalCollection>(
```
amomchilov, line 75: Adapted from https://github.com/apple/swift/blob/d6f9401/stdlib/public/core/SequenceAlgorithms.swift#L235-L252

Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,84 @@
...
+    with possibleSuffix: PossibleSuffix
+  ) -> Bool where PossibleSuffix.Element == Element {
+    return self.ends(with: possibleSuffix, by: ==)
+  }
+
+  /// Returns a Boolean value indicating whether the final elements of the
+  /// collection are equivalent to the elements in another collection, using
+  /// the given predicate as the equivalence test.
+  ///
+  /// The predicate must be a *equivalence relation* over the elements. That
```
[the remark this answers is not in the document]
amomchilov: Not from `swift-algorithms`, but in the standard library itself.

I opened a fix: https://github.com/apple/swift/pull/71871/files

Sources/Algorithms/EndsWith.swift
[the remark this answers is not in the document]
amomchilov: I'm good with either Arrays or Strings. Perhaps ranges are a bad example, given that some readers might confuse `..<` and `...`. We can pick a more self-evident example that doesn't need to require knowledge of that distinction.

Sources/Algorithms/EndsWith.swift
amomchilov: I agree.

For completeness, I'll mention an alternative implementation I toyed with, which has different constraints:

1. `self` needs to be `RandomAccessCollection` instead of `Bidirectional` (and it does some indexing offsetting, which can be slow if that's not `O(1)`).
2. The `possiblePrefix` only needs to be a `Collection` instead of a `BidirectionalCollection`

```swift
extension RandomAccessCollection where Element: Equatable {
  @inlinable
  public func ends<PossibleSuffix: Collection>(
    with possibleSuffix: PossibleSuffix,
    by areEquivalent: (Element, PossibleSuffix.Element) throws -> Bool
  ) rethrows -> Bool {
    guard let startOfSuffix = self.index(self.endIndex, offsetBy: -possibleSuffix.count, limitedBy: self.startIndex) else {
      return false
    }

    return try self[startOfSuffix...].elementsEqual(possibleSuffix, by: areEquivalent)
  }
}
```

Guides/EndsWith.md
amomchilov: I was thinking that perhaps this can be the first of several methods that fall under a new "predicates" category, which would contain boolean-returning methods which check for particular qualities in a collection.

One other predicate I've been thinking of is something like `someArray.allElementsAreEqual()`, which e.g. would be true for `[1, 1]` but not `[1, 2]`.

If we find other predicates to add to `swift-algorithms`, we can group `ends(with:)` among them.

Until then, I'm not sure which "guide" we should document this in.

Sources/Algorithms/EndsWith.swift
```diff
@@ -0,0 +1,82 @@
...
+  /// - Returns: `true` if the initial elements of the collection are equivalent
+  ///   to the elements of `possibleSuffix`; otherwise, `false`. If
+  ///   `possibleSuffix` has no elements, the return value is `true`.
+  ///
+  /// - Complexity: O(*m*), where *m* is the lesser of the length of the
+  ///   collection and the length of `possibleSuffix`.
+  @inlinable
+  public func ends<PossibleSuffix: BidirectionalCollection>(
+    with possibleSuffix: PossibleSuffix,
+    by areEquivalent: (Element, PossibleSuffix.Element) throws -> Bool
```
LucianoPAlmeida, line 77: Since equivalence is dictated by `areEquivalent` does this need to be constrained to element is `Equatable`(`BidirectionalCollection where Element: Equatable`)?

amomchilov: Nice catch! I moved this overload into its own unconstrained extension.

https://github.com/apple/swift-algorithms/compare/17dfa21b45e1d3fdfd31e9f5e3a2c361f4a7cef6..f49094216b7f225dda7ee54ec7ad6c5801dfb75c

Sources/Algorithms/EndsWith.swift
LucianoPAlmeida: Just a small addition here is that in this case `possibleSuffix` probably will be better if constrained to `RandomAccessCollection` as well, otherwise `-possibleSuffix.count` is not guaranteed to be O(1).

LucianoPAlmeida (approved):

amomchilov:
@natecook1000 Is this looking good to merge?

Naveen-C-Ramachandrappa (approved):

natecook1000 (approved):

natecook1000:
@swift-ci Please test

PR #225, opened 2024-02-29
Remove `@_implementationOnly` annotations

marcosgriselli:
These annotations produce warnings when compiling without library evolution using Swift ≥5.10.

Replace them by private import when compiling using Swift ≥5.11.

Mirrors apple/swift-syntax#2429

### Checklist

### DIFF
--- a/Sources/Algorithms/RandomSample.swift
+++ b/Sources/Algorithms/RandomSample.swift
@@ -10,8 +10,14 @@
 //===----------------------------------------------------------------------===//
 
 // For log(_:) and root(_:_:)
+#if swift(>=5.11)
+internal import RealModule
+#elseif swift(>=5.10)
+import RealModule
+#else
 @_implementationOnly
 import RealModule
+#endif
 
 //===----------------------------------------------------------------------===//
 // randomStableSample(count:)


marcosgriselli:
@natecook1000 sorry for the ping, can we add a reviewer for this?

LucianoPAlmeida:
@swift-ci Please test

natecook1000:
Looks great, thanks @marcosgriselli!

Sources/Algorithms/RandomSample.swift
```diff
@@ -10,8 +10,14 @@
 //===----------------------------------------------------------------------===//
 
 // For log(_:) and root(_:_:)
+#if swift(>=5.11)
+internal import RealModule
+#elseif swift(>=5.10)
+import RealModule
```
tshortli, line 16: Importing `RealModule` without `@_implementationOnly` in Swift 5.10, but with `@_implementationOnly` in Swift 5.9 doesn't make sense to me. With `@_implementationOnly` the dependency is hidden from dependents entirely. Without it, the dependency is exposed to dependents. If in practice it is ok to expose the `RealModule` dependency to dependents (as it seems it must be) then it would be more consistent to simply remove `@_implementationOnly` for <= 5.10, rather than having different semantics for 5.10 and <= 5.9.

PR #228, opened 2024-04-19
Clarify the docs for `partitioned(by:)`

j-f1:
As currently written, the summary for `partitioned(by:)` heavily implies the return value is `(trueValues, falseValues)`. In fact, it’s the opposite. I rephrased this summary to clarify that (1) the order is `(falseValues, trueValues)` and (2) the elements of the two returned arrays are in the order they were in the original sequence/collection.

### Checklist
- (N/A) I've added at least one test that validates that my change is working, if appropriate

### DIFF
--- a/Sources/Algorithms/Partition.swift
+++ b/Sources/Algorithms/Partition.swift
@@ -256,7 +256,7 @@ extension Sequence {
 extension Collection {
   /// Returns two arrays containing the elements of the sequence that
   /// don’t and do satisfy the given predicate, respectively. The order of the
-  /// arrays matches the order of the elements in the original sequence.
+  /// arrays matches the order of the elements in the original collection.
   ///
   /// In this example, `partitioned(by:)` is used to separate the input based on
   /// whether a name is shorter than five characters.
--- a/Sources/Algorithms/Partition.swift
+++ b/Sources/Algorithms/Partition.swift
@@ -210,8 +210,9 @@ extension Collection {
 //===----------------------------------------------------------------------===//
 
 extension Sequence {
-  /// Returns two arrays containing, in order, the elements of the sequence that
-  /// do and don’t satisfy the given predicate.
+  /// Returns two arrays containing the elements of the sequence that
+  /// don’t and do satisfy the given predicate, respectively. The order of the
+  /// arrays matches the order of the elements in the original sequence.
   ///
   /// In this example, `partitioned(by:)` is used to separate the input based on
   /// whether a name is shorter than five characters:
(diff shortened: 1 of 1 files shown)


j-f1:
@swift-ci please test

Sources/Algorithms/Partition.swift
```diff
@@ -210,8 +210,9 @@ extension Collection {
 //===----------------------------------------------------------------------===//
 
 extension Sequence {
-  /// Returns two arrays containing, in order, the elements of the sequence that
-  /// do and don’t satisfy the given predicate.
+  /// Returns two arrays containing the elements of the sequence that
+  /// don’t and do satisfy the given predicate, respectively. The order of the
```
mdznr, line 214: “The order of the arrays” makes it seem like the order of the tuple’s elements. Perhaps “in” should be used here?

```suggestion
  /// don’t and do satisfy the given predicate, respectively. The order in the
```

Sources/Algorithms/Partition.swift
```diff
@@ -253,8 +254,9 @@ extension Sequence {
 }
 
 extension Collection {
-  /// Returns two arrays containing, in order, the elements of the collection
-  /// that do and don’t satisfy the given predicate.
+  /// Returns two arrays containing the elements of the sequence that
+  /// don’t and do satisfy the given predicate, respectively. The order of the
```
mdznr, line 258:
```suggestion
  /// don’t and do satisfy the given predicate, respectively. The order in the
```

Sources/Algorithms/Partition.swift
j-f1: Adjusted the wording a bit, good point!

j-f1:
@swift-ci please test

mdznr (approved):

LucianoPAlmeida (approved):

Sources/Algorithms/Partition.swift
```diff
@@ -210,8 +210,10 @@ extension Collection {
 //===----------------------------------------------------------------------===//
 
 extension Sequence {
-  /// Returns two arrays containing, in order, the elements of the sequence that
-  /// do and don’t satisfy the given predicate.
+  /// Returns two arrays containing the elements of the sequence that
+  /// don’t and do satisfy the given predicate, respectively. The order of the
```
natecook1000, line 214: The abstract should be a single sentence. Could you move the second sentence down to the `Returns:` section?

natecook1000 (commented):
One quick note (that applies to both changes), then we should be good to merge! Thanks, @j-f1!

j-f1:
@swift-ci please test

PR #234, opened 2024-07-13
Detection of One Sorted Sequence Being a Subset of Another

CTMacUser:
### Description

This function is an adaptation of the [`includes`](https://en.cppreference.com/w/cpp/algorithm/includes) function from C++. This pull request is a sequel to ["Detection of How One Sorted Sequence Includes Another"](https://github.com/apple/swift-algorithms/pull/38) (#38).

I started with a new version of the `Inclusion` type and the `sortedOverlap` function from Pull #38. I added another function, `includes`, that calls the overlap detector then translates the exact overlap degree to a simple `Bool` result.

Then I figured that no one actually cares about the precise overlapping factor. So the precise-overlap function and support type were removed, and its code was moved into `includes` directly. Since a general answer wasn't required, I could add short-circuit logic.

### Detailed Design

One primary function extends `Sequence` to test if a given sequence is a subset of the receiver, assuming both are sorted according to the given predicate. (The given sequence's elements need not be contiguous in the receiver.) The variant function removes the predicate parameter for a default of the less-than operator (`<`), at the cost of requiring `Comparable` conformance.

```swift
extension Sequence {
    public func
    includes<T: Sequence>(sorted other: T, sortedBy areInIncreasingOrder: (Element, Element) throws -> Bool) rethrows -> Bool
    where T.Element == Element
}

    @inlinable public func
    includes<T: Sequence>(sorted other: T) -> Bool
    where T.Element == Element
}
```

### Documentation Plan

Both introductory documentation and a guide were added.

### Test Plan

A test file was added.

### Source Impact

The functions are an additive change, not otherwise affecting the API.

### Checklist

### DIFF
--- a/Guides/Inclusion.md
+++ b/Guides/Inclusion.md
@@ -1,0 +1,74 @@
+#  Inclusion
+
+[[Source](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Includes.swift) | 
+ [Tests](https://github.com/apple/swift-algorithms/blob/main/Tests/SwiftAlgorithmsTests/IncludesTests.swift)]
+
+Check if one sequence includes another.
+Both sequences must be sorted along the same criteria,
+which must provide a strict weak ordering.
+
+```swift
+let first = [4, 3, 2, 1], second = [3, 2, 1, 0], third = [3, 2]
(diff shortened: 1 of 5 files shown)


CTMacUser:
I made a [discussion thread](https://forums.swift.org/t/idea-testing-set-inclusion/73613).

Issue #245, opened 2024-11-02
Warning for Documentation.docc: found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target


atacan:
Building a swift package that has swift-algorithms as dependency outputs warning in the terminal

**Swift Algorithms version:** `1.2.0`
**Swift version:**

```
swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Target: arm64-apple-macosx15.0
```

### Checklist
- [ ] If possible, I've reproduced the issue using the `main` branch of this package
- [x] I've searched for [existing GitHub issues](https://github.com/apple/swift-algorithms/issues)

### Steps to Reproduce

Build the Hummingbird template project https://github.com/hummingbird-project/template

### Expected behavior
No warning for package documentation

### Actual behavior
```console
warning: 'swift-algorithms': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/atacan/Developer/Repositories/speech-router-hb/.build/checkouts/swift-algorithms/Sources/Algorithms/Documentation.docc
```

dimitribouniol:
Just chiming in to say I also get this on everyone of my builds and was about to file an issue 😛

fpseverino:
See https://github.com/apple/swift-algorithms/pull/247#issuecomment-2599683577

PR #247, opened 2024-12-17
Exclude `docc` files from sources to avoid build warnings

omochi:
When using this package as a dependency, the following warning appears during the build process:

```
warning: 'swift-algorithms': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /Users/omochi/<myproject>/.build/checkouts/swift-algorithms/Sources/Algorithms/Documentation.docc
```

This happens because `docc` files are treated as undefined source files.
This patch explicitly excludes them using the `exclude` directive, clarifying that they are not source files and preventing the warning from being displayed.

**Relations**
- #245

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,7 +27,8 @@ let package = Package(
             name: "Algorithms",
             dependencies: [
               .product(name: "RealModule", package: "swift-numerics"),
-            ]),
+            ],
+            exclude: ["Documentation.docc"]),
         .testTarget(
             name: "SwiftAlgorithmsTests",
             dependencies: ["Algorithms"]),


fpseverino:
This warning is thrown when `swift-tools-version` is 5.4 or lower, the `main` branch is now on 5.7, but the latest release is still on 5.4.

So this change isn't really necessary, a new release should fix the problem.

omochi:
I experimented locally, and indeed, no warnings about DocC appeared on the main branch.
I hope version 1.2.1 or 1.3 will be released soon.

PR #256, opened 2025-04-16
Partition map

iDmitriyy:
MR is made as the result of the discussion with @natecook1000 : https://forums.swift.org/t/add-splitmap-function-to-swift-standard-library/69321

### Checklist


iDmitriyy:
One thing needed to discuss. As we decided to not make internal Either enum public, current names a PartitionMapResult2 and PartitionMapResult3.
While these types are made for concrete purpose, names are felt to be very long. My current thoughts are to use shorter names like OneOf2Result / OneOf3Result.

PR #259, opened 2025-08-08
Add labels to the tuples produced by `chunked(on:)`

cweider:
Label the tuple produced by `chunked(on:)` with `subject` and `chunk`. This will improve the legibility of code downstream of the said function. For example, when providing an identifier for SwiftUI's `ForEach`:

```
-ForEach(chunkedByDate, id: \.0) { date, items in
+ForEach(chunkedByDate, id: \.subject) { date, items in
```

This aligns `ChunkedOnCollection` with the `IndexedCollection`, which specifies `index` and `element` as well as the built-in `EnumeratedSequence` (specifying `offset` and `element`).

The introduction of the label will have no effect on existing code, because the index-based address of the tuple component will remain unchanged and absent labels will be inferred. This said, the label choice should be considered carefully since subsequent relabelings will disrupt established code.

### Checklist
- [ ] I've read the [Contribution Guidelines](../blob/main/CONTRIBUTING.md)

### DIFF
--- a/Sources/Algorithms/Chunked.swift
+++ b/Sources/Algorithms/Chunked.swift
@@ -162,6 +162,7 @@ public struct ChunkedOnCollection<Base: Collection, Subject: Equatable> {
 
 extension ChunkedOnCollection: Collection {
   public typealias Index = ChunkedByCollection<Base, Subject>.Index
+  public typealias Element = (subject: Subject, chunk: Base.SubSequence)
 
   @inlinable
   public var startIndex: Index {
@@ -174,7 +175,7 @@ extension ChunkedOnCollection: Collection {
   }
 
   @inlinable
-  public subscript(position: Index) -> (Subject, Base.SubSequence) {
+  public subscript(position: Index) -> Element {
     let subsequence = chunked[position]
     // swift-format-ignore: NeverForceUnwrap
     // Chunks are never empty, so `.first!` is safe.
@@ -509,7 +510,7 @@ extension Collection {
   @inlinable
   public func chunked<Subject: Equatable>(
     on projection: (Element) throws -> Subject
-  ) rethrows -> [(Subject, SubSequence)] {
+  ) rethrows -> [ChunkedOnCollection<Self, Subject>.Element] {
     guard !isEmpty else { return [] }
     var result: [(Subject, SubSequence)] = []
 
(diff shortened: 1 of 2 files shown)


cweider:
Aside: Ruby’s analogue, [`chunk`](https://github.com/ruby/ruby/blob/07878eb/enum.c#L3881), has no meaningful naming to take guidance from.

Tests/SwiftAlgorithmsTests/ChunkedTests.swift
```diff
@@ -71,6 +71,16 @@ final class ChunkedTests: XCTestCase {
...
 
+  func testChunkedOnLabels() {
+    let arrayChunks: Array = fruits.chunked(on: { $0.first })
+    XCTAssert(arrayChunks.first!.0 == arrayChunks.first!.subject)
+    XCTAssert(arrayChunks.first!.1 == arrayChunks.first!.chunk)
+
+    let lazyChunks = fruits.lazy.chunked(on: { $0.first })
+    XCTAssert(lazyChunks.first!.0 == lazyChunks.first!.subject)
+    XCTAssert(lazyChunks.first!.1 == lazyChunks.first!.chunk)
+  }
```
cweider, line 82: Test is a bit artificial – correctness is known statically. The best justification is that inhibits subsequent change to the labels. I welcome your thoughts w.r.t. testing.

Sources/Algorithms/Chunked.swift
```diff
@@ -509,7 +510,7 @@ extension Collection {
   @inlinable
   public func chunked<Subject: Equatable>(
     on projection: (Element) throws -> Subject
-  ) rethrows -> [(Subject, SubSequence)] {
+  ) rethrows -> [ChunkedOnCollection<Self, Subject>.Element] {
```
cweider, line 513: The implementation below sidesteps `ChunkedOnCollection`, but aliasing seems sensible given `Array`-ness is the only intended difference between this and the `LazySequenceProtocol` function.

Sources/Algorithms/Chunked.swift
```diff
@@ -162,6 +162,7 @@ public struct ChunkedOnCollection<Base: Collection, Subject: Equatable> {
 
 extension ChunkedOnCollection: Collection {
   public typealias Index = ChunkedByCollection<Base, Subject>.Index
+  public typealias Element = (subject: Subject, chunk: Base.SubSequence)
```
cweider, line 165: cf. `IndexedCollection`:

https://github.com/apple/swift-algorithms/blob/0b70cac980199f99ced9e4f03530f28d011b23a1/Sources/Algorithms/Indexed.swift#L25-L27

xwu:
> The introduction of the label will have no effect on existing code

This is not strictly true because of Swift's tuple shuffle behavior, although it is unlikely to break correct working code.

